PR #1, opened 2019-04-30
Add docker compose wrapper for install, start, prune, stop commands

se7entyse7en:
This is the simplest possible wrapper around docker-compose for handling the lifecycle of the product.


smacker:
though if I was writing it I would put everything in 1 file.

cmd/sandbox-ce/cmd/prune.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/smacker/superset-compose/cmd/sandbox-ce/compose"
+)
+
+type pruneCmd struct {
+	Command `name:"prune" short-description:"Prune"`
+}
+
+func (c *pruneCmd) Execute(args []string) error {
+	return compose.Run(context.Background(), "down")
```
smacker: most probably we want `--volumes` here

se7entyse7en: Addressed [here](https://github.com/smacker/superset-compose/pull/1/commits/54eab1be637ae71878dd9b31221d709de3632dc5).

PR #2, opened 2019-04-30
Bblfsh web

smacker:
This PR adds bblfsh-web into superset.

It also contains simple Makefile to apply changes on superset.

*Note*:
It is **not** the way to go for us, in my opinion.
I have just copied original files to see the diff for now.
It means updates for that files must be done manually.
In theory, we can do something with `git filter-branch` or `git checkout -- path` with another remote or somehow else implement `git merge` for those files only.
Though it's out of the scope of this PR.

It looks like this after `make build` && `docker-compose up`
Screenshot 2019-04-30 at 17 54 43


se7entyse7en:
It looks really good!

Makefile
```diff
@@ -1,7 +1,37 @@
...
+		cp -r "$(PATCH_SOURCE_DIR)/$${file}" "$(SUPERSET_DIR)/$${file}"; \
+	done; \
+
+# Overrides files in the superset repository using symlinks
+.PHONY: patch-dev
+patch-dev:
+	@for file in $(ADD_FILES) $(OVERRIDE_FILES); do \
+		echo "patching $${file}"; \
+		rm -rf "$(SUPERSET_DIR)/$${file}"; \
+		ln -s "$(PATCH_SOURCE_DIR)/$${file}" "$(SUPERSET_DIR)/$${file}"; \
```
se7entyse7en: Why `cp` in one case and `ln` in the other?

smacker: we must `cp` to build docker image. Or `build` will fail.

`ln` is for development. So you can run `FLASK_ENV=development flask run -p 8088 --with-threads --reload --debugger` with `npm run dev-server` in `superset` dir locally and have auto-reload when you change the files.

se7entyse7en: Great! Thanks for the explanation

PR #3, opened 2019-04-30
pin superset on 0.32 release branch

smacker:
Master already has some breaking changes.


smacker:
@carlosms are you okay with this version?

smacker:
we discovered a bug. I'll pin version in bbfsh-brunch and fix the bug.

PR #4, opened 2019-04-30
Add fallback to docker-compose running inside container

se7entyse7en:
This depends on #1.

If `docker-compose` is not installed in the host, then we fallback to `docker-compose` run inside docker container. The utility binary that runs `docker-compose` inside a container is saved into `~/.srcd/bin`.

### DIFF
--- a/cmd/sandbox-ce/cmd/install.go
+++ b/cmd/sandbox-ce/cmd/install.go
@@ -19,7 +19,6 @@ func (c *installCmd) Execute(args []string) error {
 	}
 
 	err = compose.Run(context.Background(), "up")
-
 	if err != nil {
 		return err
 	}
--- a/cmd/sandbox-ce/compose/compose.go
+++ b/cmd/sandbox-ce/compose/compose.go
@@ -19,9 +19,7 @@ import (
 // that it's not already present in the system
 const composeContainerPath = "https://github.com/docker/compose/releases/download/1.24.0/run.sh"
 
-var envKeys = map[string]bool{
-	"GITBASE_REPOS_DIR": true,
-}
+var envKeys = []string{"GITBASE_REPOS_DIR"}
 
 type Compose struct {
 	bin string
@@ -31,7 +29,7 @@ func (c *Compose) Run(ctx context.Context, arg ...string) error {
 	cmd := exec.CommandContext(ctx, c.bin, arg...)
 
 	var compOpts []string
-	for key := range envKeys {
+	for _, key := range envKeys {
 		val, ok := os.LookupEnv(key)
 		if !ok {
 			continue
--- a/cmd/sandbox-ce/compose/compose.go
+++ b/cmd/sandbox-ce/compose/compose.go
@@ -14,6 +14,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+// composeContainerPath is the url docker-compose is downloaded from in case
+// that it's not already present in the system
 const composeContainerPath = "https://github.com/docker/compose/releases/download/1.24.0/run.sh"
 
 var envKeys = map[string]bool{
--- a/cmd/sandbox-ce/compose/compose.go
+++ b/cmd/sandbox-ce/compose/compose.go
@@ -40,9 +40,9 @@ func (c *Compose) Run(ctx context.Context, arg ...string) error {
 		compOpts = append(compOpts, fmt.Sprintf("-e %s=%s", key, val))
 	}
 
-	cmd.Env = []string{
-		fmt.Sprintf("COMPOSE_OPTIONS=%s", strings.Join(compOpts, " ")),
-	}
+	cmd.Env = append(os.Environ(),
+		fmt.Sprintf("COMPOSE_OPTIONS=%s", strings.Join(compOpts, " ")))
+
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
--- a/cmd/sandbox-ce/compose/compose.go
+++ b/cmd/sandbox-ce/compose/compose.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/mitchellh/go-homedir"
@@ -108,6 +109,10 @@ func getOrInstallComposeContainer() (string, error) {
 }
 
 func downloadCompose(path string) error {
+	if runtime.GOOS == "windows" {
+		return fmt.Errorf("compose in container is not compatible with Windows")
+	}
+
 	out, err := os.Create(path)
 	if err != nil {
 		return err
--- a/cmd/sandbox-ce/compose/compose.go
+++ b/cmd/sandbox-ce/compose/compose.go
@@ -2,16 +2,133 @@ package compose
 
 import (
 	"context"
+	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/pkg/errors"
 )
 
-func Run(ctx context.Context, arg ...string) error {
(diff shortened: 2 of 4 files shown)


se7entyse7en:
I removed `docker-compose` locally and it woked really smooth, I only had to propagate the env vars to docker.

cmd/sandbox-ce/compose/compose.go
```diff
@@ -2,16 +2,117 @@ package compose
...
+
+type Compose struct {
+	bin string
+}
+
+func (c *Compose) Run(ctx context.Context, arg ...string) error {
+	cmd := exec.CommandContext(ctx, c.bin, arg...)
+
+	var compOpts []string
+	for _, keyValue := range os.Environ() {
```
smacker: why no [LookupEnv](https://golang.org/pkg/os/#LookupEnv)?

se7entyse7en: Simply because I didn't know it 😂
Thanks for the tip!

cmd/sandbox-ce/compose/compose.go
```diff
@@ -2,16 +2,117 @@ package compose
...
+		return "", err
+	}
+
+	err = os.MkdirAll(dirPath, os.ModePerm)
+	if err != nil {
+		return "", fmt.Errorf("error while creating directory %s", dirPath)
+	}
+
+	cmd := exec.CommandContext(context.Background(),
+		"curl", "-L", "--fail", composeContainerPath, "-o", path)
```
smacker: It adds curl dependency that can be missed on the host

se7entyse7en: that's true, maybe better replace with `net/http` and download.

se7entyse7en: Addressed [here](https://github.com/smacker/superset-compose/pull/4/commits/a42783d7aee883af91e9ea9836fd56eaa3eb6474).

cmd/sandbox-ce/compose/compose.go
se7entyse7en: Addressed [here](https://github.com/smacker/superset-compose/pull/4/commits/5646ab04f2b558d5bb64d139716bb39e95476f70).

cmd/sandbox-ce/cmd/install.go
```diff
@@ -0,0 +1,32 @@
...
+func (c *installCmd) Execute(args []string) error {
+	err := compose.Run(context.Background(),
+		"run", "--rm", "superset", "./docker-init.sh")
+
+	if err != nil {
+		return err
+	}
+
+	err = compose.Run(context.Background(), "up")
+
```
smacker: just a nit pick. But usually we don't put empty line in such cases...

cmd/sandbox-ce/compose/compose.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func getOrInstallComposeContainer() (string, error) {
+	homedir, err := homedir.Dir()
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to get home dir")
+	}
+
+	dirPath := filepath.Join(homedir, ".srcd", "bin")
+	path := filepath.Join(dirPath, "docker-compose")
```
smacker: I don't know if it's a problem or not. But if you change the version of the script above it will still use the old one.

se7entyse7en: I think that it's better to keep a fixed version and to update the `composeContainerPath` manually when we need so that we're sure about the expected behavior. On the other hand, we don't have this fully controlled as the user could have its own docker-compose already installed.

So yup, we could eventually always check for update. But in case I'd add this in a second iteration.

se7entyse7en:
@carlosms yup, that was actually a bug in the branch this was rebased on. It's fixed now, I'm gonna rebase this on master and then it should work.

cmd/sandbox-ce/cmd/install.go
se7entyse7en: fixed

cmd/sandbox-ce/compose/compose.go
se7entyse7en: 🤦‍♂ By reading again at this I understood what you mean. I'll add a comment that the version is in the `composeContainerPath` is considered only the first time, but it will be ignored if the script is already present. I'm also gonna open an issue for this.

se7entyse7en: Addressed [here](https://github.com/smacker/superset-compose/pull/4/commits/5159a41e5d0ac7a749a48df7c9672f078d3fcea6).

se7entyse7en:
@carlosms sorry I was wrong, that was actually a bug that should be present in `master`, but that should have been fixed in this PR. In `master` there's no propagation of the env vars to the container.

se7entyse7en:
@carlosms I've just rebased this on latest `master`. Can you try again whether it still gives you problems?

carlosms:
Exact same problem :/

cmd/sandbox-ce/compose/compose.go
```diff
@@ -2,16 +2,135 @@ package compose
...
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/pkg/errors"
 )
 
-func Run(ctx context.Context, arg ...string) error {
-	cmd := exec.CommandContext(ctx, "docker-compose", arg...)
+// composeContainerPath is the url docker-compose is downloaded from in case
+// that it's not already present in the system
+const composeContainerPath = "https://github.com/docker/compose/releases/download/1.24.0/run.sh"
```
carlosms: If this is a bash, we should have some checks and avoid to download and run it if the os is windows.

cmd/sandbox-ce/compose/compose.go
```diff
@@ -2,16 +2,135 @@ package compose
...
+	for key := range envKeys {
+		val, ok := os.LookupEnv(key)
+		if !ok {
+			continue
+		}
+
+		compOpts = append(compOpts, fmt.Sprintf("-e %s=%s", key, val))
+	}
 
+	cmd.Env = []string{
```
carlosms: To make the command work in my linux machine, instead of replacing the `cmd` environment I changed this to append to a copy of the current process env, like this:
```go
cmd.Env = append(os.Environ(), fmt.Sprintf("COMPOSE_OPTIONS=%s", strings.Join(compOpts, " ")))
```

cmd/sandbox-ce/compose/compose.go
```diff
@@ -2,16 +2,135 @@ package compose
...
+	"github.com/pkg/errors"
 )
 
-func Run(ctx context.Context, arg ...string) error {
-	cmd := exec.CommandContext(ctx, "docker-compose", arg...)
+// composeContainerPath is the url docker-compose is downloaded from in case
+// that it's not already present in the system
+const composeContainerPath = "https://github.com/docker/compose/releases/download/1.24.0/run.sh"
+
+var envKeys = map[string]bool{
```
carlosms: Why is this a `map`? It looks like it could simply be a `[]string`

cmd/sandbox-ce/compose/compose.go
se7entyse7en: Addressed [here](https://github.com/smacker/superset-compose/pull/4/commits/b27087441b3db77b98c4b0a648122006858b582e).

cmd/sandbox-ce/compose/compose.go
se7entyse7en: Addressed [here](https://github.com/smacker/superset-compose/pull/4/commits/4956df0db7336fed033bf17058822f8ca96c4560).

Thanks for this! Today the same error that was occurring to you was happening to me, don't know why maybe I was always using the same shell with some exported env vars.

cmd/sandbox-ce/compose/compose.go
se7entyse7en: Addressed [here](https://github.com/smacker/superset-compose/pull/4/commits/5f9fc1a92dc709c2dfea3969dbc7b85542c7de66).

PR #6, opened 2019-05-01
add gitbase tables

smacker:
It fails to fetch columns for `blobs`, `files`, `ref_commits` tables.
Gitbase issue: https://github.com/src-d/gitbase/issues/808

### DIFF
--- a/docker/add_gitbase.py
+++ b/docker/add_gitbase.py
@@ -1,5 +1,6 @@
 from superset import conf, db
 from superset.models import core as models
+from superset.connectors.connector_registry import ConnectorRegistry
 
 
 def get_or_create_gitbase_db():
@@ -17,5 +18,32 @@ def get_or_create_gitbase_db():
     db.session.add(dbobj)
     db.session.commit()
 
+    return dbobj
+
+
+def add_gitbase_tables():
+    dbobj = get_or_create_gitbase_db()
+    TBL = ConnectorRegistry.sources['table']
+    for table in dbobj.all_table_names_in_schema('gitbase'):
+        # table_name should match the one in the datasource for fetch_metadata to work
+        if db.session.query(TBL).filter_by(table_name=table).first():
+            continue
+        if db.session.query(TBL).filter_by(table_name='gitbase.' + table).first():
+            continue
+
+        # create table with original name and fetch columns
+        tbl = TBL(table_name=table)
+        tbl.database = dbobj
+        db.session.add(tbl)
+        db.session.commit()
+        tbl.fetch_metadata()
+
+        # rename with prefix and set source
+        tbl.table_name = 'gitbase.' + table
+        tbl.sql = 'select * from ' + table
+        db.session.add(dbobj)
+        db.session.commit()
+
 
 get_or_create_gitbase_db()
+add_gitbase_tables()


docker/add_gitbase.py
```diff
@@ -17,5 +18,32 @@ def get_or_create_gitbase_db():
     db.session.add(dbobj)
```
se7entyse7en: Shouldn't this be added and committed only `if not dbobj`?

docker/add_gitbase.py
```diff
@@ -17,5 +18,32 @@ def get_or_create_gitbase_db():
...
+def add_gitbase_tables():
+    dbobj = get_or_create_gitbase_db()
+    TBL = ConnectorRegistry.sources['table']
+    for table in dbobj.all_table_names_in_schema('gitbase'):
+        # table_name should match the one in the datasource for fetch_metadata to work
+        if db.session.query(TBL).filter_by(table_name=table).first():
+            continue
+        if db.session.query(TBL).filter_by(table_name='gitbase.' + table).first():
+            continue
+
```
se7entyse7en: I don't understand, so we have to tables for each `gitbase` table? One with `gitbase.` prefix and one without?

docker/add_gitbase.py
```diff
@@ -17,5 +18,32 @@ def get_or_create_gitbase_db():
...
+        # create table with original name and fetch columns
+        tbl = TBL(table_name=table)
+        tbl.database = dbobj
+        db.session.add(tbl)
+        db.session.commit()
+        tbl.fetch_metadata()
+
+        # rename with prefix and set source
+        tbl.table_name = 'gitbase.' + table
+        tbl.sql = 'select * from ' + table
```
se7entyse7en: Does this create a sort of view of the table? For example, can I already hardcode a `WHERE` or `LIMIT` clause?

se7entyse7en:
Can you also clarify how does this relate to https://github.com/src-d/gitbase/issues/808?

docker/add_gitbase.py
smacker: no. Because it updates `GITBASE_DATABASE_URI`.
I might do extra update queries but we run it only once on init so it's not a problem.
Also it works the same as superset's data example scripts.

docker/add_gitbase.py
smacker: No, we have only one with prefix.
But to create it we first need to create it without prefix and then rename.

Explanation why: https://src-d.slack.com/archives/C7TB5NEDN/p1556710827229300

docker/add_gitbase.py
smacker: yes. In supersets "views" are called "tables".
You can add WHERE, LIMIT, custom columns (for example `LANGUAGE(filename, content)`) and whatever else.

smacker:
`tbl.fetch_metadata()` makes queries to datasource to get table columns and save those columns into DB (each table in superset must have columns to operate with it).

The main query that sqalchemy uses for that is `SHOW CREATE TABLE` which currently returns an incorrect result for 3 tables mentioned above.

docker/add_gitbase.py
se7entyse7en: Thanks!

PR #16, opened 2019-05-02
Superset subtree

smacker:
Use subtree merging for superset + copy for new files.

I also want to try using a separate package instead of copying files.


Makefile
```diff
@@ -1,42 +1,62 @@
...
-# Overrides files in the superset repository using symlinks
+# Copy src-d files in the superset repository using symlinks. it's useful for development.
+# Allows to run flask locally and work only inside superset directory.
 .PHONY: patch-dev
-patch-dev:
-	@for file in $(ADD_FILES) $(OVERRIDE_FILES); do \
-		echo "patching $${file}"; \
-		rm -rf "$(SUPERSET_DIR)/$${file}"; \
-		ln -s "$(PATCH_SOURCE_DIR)/$${file}" "$(SUPERSET_DIR)/$${file}"; \
+patch-dev: clean
```
carlosms: This is not working for me:
```
ln: failed to create symbolic link 'superset/superset/assets/images\/sourced-logo-2x.png': No such file or directory
```

Why don't we use this instead?
```
patch-dev: clean
	cp --symbolic-link -r $(PWD)/$(PATCH_SOURCE_DIR)/* $(PWD)/$(SUPERSET_DIR)/
	ln -s "$(PWD)/$(PATCH_SOURCE_DIR)/superset/superset_config_dev.py" "$(SUPERSET_DIR)/superset_config.py"; \
```

smacker: > This is not working for me

could you please show results or the diff and awk? On the first look, the problem is with escaping `/`.

> Why don't we use this instead?

such argument doesn't exist on macOs.

```
cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
```

carlosms:
```
diff -r srcd superset | grep "srcd"
Only in srcd/superset/assets/images: sourced-logo-2x.png
Only in srcd/superset/assets/src: uast
Only in srcd/superset: bblfsh
Only in srcd/superset: superset_config_dev.py
Only in srcd/superset: superset_config.py

diff -r srcd superset | grep "srcd" | awk '{gsub(/: /,"\/");print $3}'
srcd/superset/assets/images\/sourced-logo-2x.png
srcd/superset/assets/src\/uast
srcd/superset\/bblfsh
srcd/superset\/superset_config_dev.py
srcd/superset\/superset_config.py
```

smacker: fixed. On macOs it works both with and without escaping.

PR #27, opened 2019-05-06
Add build target for `cmd/sandbox-ce` command

### LINKED ISSUE #10: Build system for sandbox-ce command
This issue is to add build scripts for the command introduced in #1.
The default option is to use https://github.com/src-d/ci, but I don't think this is a hard requirement.

se7entyse7en:
Closes #10.


go.sum
```diff
@@ -1,6 +1,7 @@
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/jessevdk/go-flags v1.4.0 h1:4IU2WS7AumrZ/40jfhf4QVDMsQwqA7VEHozFRrGARJA=
 github.com/jessevdk/go-flags v1.4.0/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
+github.com/konsorten/go-windows-terminal-sequences v1.0.1 h1:mweAR1A6xJ3oS2pRaGiHgQ4OO8tzTaLawm8vnODuwDk=
```
smacker: why did you have to update it in this PR? It doesn't touch any go code, only Makefile.

se7entyse7en: this is something that is being added when building for `windows`.

PR #29, opened 2019-05-06
Add `config` command

se7entyse7en:
This is rebased on #4 and partially addresses #13.

This PR adds a `config` command with the following sub-command:
- `download`:
```
Usage:
  sandbox-ce [OPTIONS] config download [download-OPTIONS]

Help Options:
  -h, --help                                     Show this help message

[download command options]
      -r, --revision=                            revision to download (default: master)
```
Each downloaded `docker-compose.yml` will be put in `~/.srcd/composer-configs/<revision>/docker-compose.yml`.
- `list`: simply lists the available configurations, and marks the current default configuration as follows:
```
✔  se7entyse7en in ~/Projects/smacker/superset-compose (composer-config-download)  $ GITBASE_REPOS_DIR=/Users/se7entyse7en/Projects/.go-workspace/src/github.com/src-d go run cmd/sandbox-ce/main.go config list
cfbe758 (default)
master
```
- `set-default`:
```
Usage:
  sandbox-ce [OPTIONS] config set-default [set-default-OPTIONS] revision

Help Options:
  -h, --help                                     Show this help message
```
Sets the default configuration to use. Under the hood, it creates a symlink from `~/.srcd/composer-configs/__default__/docker-compose.yml` to the provided configuration.

This PR doesn't use the default config yet, but only adds the logic for the `config` command. The next step is to add a `--config` to the `init` and `start` commands. If `--config` is not provided, then the default config is used. If even the default config is not set, then the one present in the repo is used.


carlosms:
Regarding the code, looks good to me.

I have a few comments on UX.

I'm not sure about the sub command `config` name. A `docker-compose.yml` is called "Compose file" in the docker docs, not "config file". Plus in the future we may have config files for `sandbox-ce` itself (maybe for ports, default user password, etc). I propose to use `sandbox-ce compose download/list` instead, as it makes more clear that you are managing docker-compose files behind the scenes.

`download` does not output anything in case of success. I think it would be useful to print something like "Compose file successfully downloaded to _path_"

All commands are missing the help texts. In this case I think `config`/`compose` should include a mention to the `~/.srcd/composer-configs/`, to make it clear that you may want to go there to manage things manually.

Similar to the `config` sub command name, I feel the `composer-configs` dir could be renamed to `compose-files`, to be more aligned with the name used in the docker documentation.

smacker:
I actually disagree with Carlos. We try to avoid much complexity and use as much of docker-compose as possible. So such things as ports are already in docker-compose.yml and it makes sense to put login/pass there too (they need to be read by scripts so they should be env variables that are defined in yml). So I would prefer to keep the name "config".

smacker:
But I have concerns about UX. I think we need to discuss it with product or Santiago or Maximo.
I don't see much use of specifying and downloading different revisions.
For me it feels like we need just "update" command and allow to pass url for special cases.

carlosms:
> I actually disagree with Carlos. We try to avoid much complexity and use as much of docker-compose as possible. So such things as ports are already in docker-compose.yml and it makes sense to put login/pass there too (they need to be read by scripts so they should be env variables that are defined in yml). So I would prefer to keep the name "config".

The point I was trying to make is not about reserving `config` for other files. Rather, my main point is that `sandbox-ce compose` makes it clear that the user is dealing with docker compose files.

If the docs say you can update your "config", the implications are not clear. If we refer to them clearly as docker compose files, it's more easy to understand its purpose.

This goes in line with the example you propose. If we say in the docs that you can setup some environment variables in the "config" file, a user might be surprised to find a docker-compose.yml. Also, calling these files "compose files" means that a user that is not familiar with docker compose can look up reference docs for it if some customization is needed.

se7entyse7en:
We discussed in apps sync, and the final decision is [here](https://github.com/src-d/minutes/pull/577/files#diff-b8dfe09eeb40c44d88f3164628c37a01R150).

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,201 @@
...
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/mitchellh/go-homedir"
```
smacker: we don't need external package for it anymore. Go 1.12 introduced function to get home dir in stdlib. Take a look at: https://github.com/src-d/superset-compose/pull/51

se7entyse7en: done

se7entyse7en:
I rebased this and updated in order to support also urls.

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,216 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"github.com/pkg/errors"
+	"gopkg.in/src-d/go-cli.v0"
+)
+
+const orgName = "smacker"
```
smacker: it's `src-d` now

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,216 @@
...
+const composerConfigFileURL = "https://raw.githubusercontent.com/%s/%s/%s/docker-compose.yml"
+
+type configCmd struct {
+	cli.PlainCommand `name:"config" short-description:"Config"`
+}
+
+type configDownloadCmd struct {
+	Command `name:"download" short-description:"Download"`
+
+	Revision string `short:"r" long:"revision" description:"revision to download" default:"master"`
```
smacker: just a suggestion (maybe a bit controversial): instead of 2 different flags use only one that can be filled by url or by tag. It's easy to distinguish them in code.

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,216 @@
...
+
+	err = c.downloadConfig(confURL, outPath)
+	if err != nil {
+		return err
+	}
+
+	return setDefaultConfig(outPath)
+}
+
+func (c *configDownloadCmd) downloadConfig(url, path string) error {
```
smacker: Maybe we should reuse the same code for both initial download (that is already in master) and for download command.

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,216 @@
...
+	if _, err := os.Lstat(defaultConfPath); err == nil {
+		if err := os.Remove(defaultConfPath); err != nil {
+			return fmt.Errorf("failed to unlink: %+v", err)
+		}
+	}
+
+	return os.Symlink(confPath, defaultConfPath)
+}
+
+func getConfigDirPath() (string, error) {
```
smacker: similar here. Let's abstract this code to reuse in both commands. And in new "update" command also.

cmd/sandbox-ce/cmd/config.go
carlosms: Even if it's easy to code it, will it be easy to understand for the user? How would you call an option that accepts either a URL or a tag?

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		fmt.Println(entry)
+	}
+
+	return nil
+}
+
+type configSetDefaultCmd struct {
+	Command `name:"set-default" short-description:"Set default config"`
```
carlosms: Maybe `set` would be enough, to make it shorter.
Also, I think it makes more sense to refer to it as the "active" config, not the "default" config. Or "current". But this is personal preference, I don't mind too much one way or the other.

cmd/sandbox-ce/cmd/config.go
smacker: as I said it might be a controversial suggestion. In my opinion this:
```
srcd download v1.2.3
srcd download https://sourced.tech/customer/special.yml
```
is better than
```
srcd download --revision v1.2.3
srcd download --url https://sourced.tech/customer/special.yml
```

carlosms: what about the argument name for the help? "version" maybe? as in, download the _version_ in this tag, or the _version_ in this URL...

smacker: *version* is good

carlosms:
I have pushed changes addressing most comments.

Main changes:
- Moved all functions that deal with docker compose files to a new package, `compose/file`
- `download` takes an argument that can be either a revision or a URL.
- For the files downloaded from URL, I have changed the dir name from `time.Now` to the URL in base64. This way the name is safe to use as dir name, and can be decoded to use the original URL in `list`.
- I have renamed all functions that deal with docker compose files from `config` to `compose` or `compose files`.
I feel `sandbox-ce compose` conveys better than `config` what the command is about. Let me know if you have strong opinions on this, if you think it should be named back to `sandbox-ce config`.

cmd/sandbox-ce/cmd/config.go
carlosms: fixed here https://github.com/src-d/superset-compose/pull/29/files#diff-28997035031024d6feab4265612f9909R19

cmd/sandbox-ce/cmd/config.go
carlosms: Done here https://github.com/src-d/superset-compose/pull/29/files#diff-0ef40ee555d59e69cfef14c4d469b691R19

cmd/sandbox-ce/cmd/config.go
carlosms: refactored, see https://github.com/src-d/superset-compose/pull/29/files#diff-480c3c3933838ac5fb6eebb456127ab1R65

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,216 @@
...
+func (c *configDownloadCmd) downloadConfig(url, path string) error {
+	out, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+
+	defer out.Close()
+
+	resp, err := http.Get(url)
+	if err != nil {
```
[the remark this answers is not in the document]
carlosms: Now fixed here https://github.com/src-d/superset-compose/pull/29/files#diff-d8b77fd89aa3fac2ee85236c3b13f844R33

cmd/sandbox-ce/cmd/config.go
```diff
@@ -0,0 +1,216 @@
...
+	URL      string `short:"u" long:"url" description:"url of the config to download"`
+}
+
+func (c *configDownloadCmd) Execute(args []string) error {
+	var err error
+	var confURL, outPath string
+
+	if c.URL != "" {
+		confURL = c.URL
+		outPath, err = getConfigPath(strconv.Itoa(int(time.Now().Unix())))
```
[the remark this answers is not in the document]
carlosms: Now the dir name is the base64 of the URL. So the same URL will override the previous download.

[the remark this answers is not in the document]
carlosms: Now the dir name is the base64 of the URL. So the `list` command prints the decoded URL again.

carlosms:
I agree base64 is not ideal when you `cd` into the dir, but well at least `list` shows the proper name.

Maybe a simple improvement could be to add a README.txt file in the same dir and put there the clean URL. In any case I would consider this as a future improvement.

PR #31, opened 2019-05-07
Web ui opening

### LINKED ISSUE #17: Add sub command to open the web interface
From the design doc:
> As a user I want to open the web interface so that I don't need to know the exact URL/port it is running and copy and paste to my browser

I'm not sure about the implementation details, but my first proposal is to open the browser on `start`, and maybe have a new sub command (`web`?) that only does this, open the link in the browser.

se7entyse7en:
This is rebased on #4 and closes #17.

Opens the browser with web ui on `install` and `start` and adds a `web` command.

### DIFF
--- a/cmd/sandbox-ce/cmd/web.go
+++ b/cmd/sandbox-ce/cmd/web.go
@@ -14,6 +14,22 @@ import (
 	"github.com/smacker/superset-compose/cmd/sandbox-ce/compose"
 )
 
+type webCmd struct {
+	Command `name:"web" short-description:"Web"`
+}
+
+func (c *webCmd) Execute(args []string) error {
+	if err := <-OpenUI(10 * time.Second); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func init() {
+	rootCmd.AddCommand(&webCmd{})
+}
+
 func openUI() error {
 	var stdout, stderr bytes.Buffer
 	for {
--- a/cmd/sandbox-ce/cmd/web.go
+++ b/cmd/sandbox-ce/cmd/web.go
@@ -1,0 +1,70 @@
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/pkg/browser"
+	"github.com/pkg/errors"
+	"github.com/smacker/superset-compose/cmd/sandbox-ce/compose"
+)
+
+func openUI() error {
+	var stdout, stderr bytes.Buffer
+	for {
+		if err := compose.RunWithIO(context.Background(),
+			os.Stdin, &stdout, &stderr, "port", "superset", "8088"); err == nil {
+			break
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+
+	address := strings.TrimSpace(stdout.String())
+	if address == "" {
+		return fmt.Errorf("no address found")
+	}
+
+	for {
+		if _, err := net.Dial("tcp", address); err == nil {
+			break
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+
+	if err := browser.OpenURL(fmt.Sprintf("http://%s", address)); err != nil {
+		errors.Wrap(err, "cannot open browser")
+	}
+
+	return nil
+}
+
+// OpenUI opens the browser with the UI.
+//
+// If opening the UI raises an error or took more time then `timeout`, then the
+// error sent to the returned channel.
+func OpenUI(timeout time.Duration) <-chan error {
+	done := make(chan error)
+
+	go func() {
+		ch := make(chan error)
+		go func() {
+			ch <- openUI()
+		}()
+
+		select {
+		case err := <-ch:
+			done <- errors.Wrap(err, "an error occured while opening ui")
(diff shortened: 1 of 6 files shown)


cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+
+	go func() {
+		ch := make(chan error)
+		go func() {
+			ch <- openUI()
+		}()
+
+		select {
+		case err := <-ch:
+			done <- errors.Wrap(err, "an error occured while opening ui")
```
carlosms:
```suggestion
			done <- errors.Wrap(err, "an error occurred while opening the UI")
```

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+		case <-time.After(timeout):
+			done <- fmt.Errorf("opening the ui took more than %v", timeout)
```
carlosms:
```suggestion
			done <- fmt.Errorf("opening the UI took more than %v", timeout)
```

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+	for {
+		if err := compose.RunWithIO(context.Background(),
+			os.Stdin, &stdout, &stderr, "port", "superset", "8088"); err == nil {
+			break
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+
+	address := strings.TrimSpace(stdout.String())
```
carlosms: For me this is opening `http://0.0.0.0:8088`, which feels very wrong as it is not a valid destination IP, and I'm still surprised every time I realize chrome treats it like localhost. Maybe we could replace `0.0.0.0` with `127.0.0.1`?

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+	if err := browser.OpenURL(fmt.Sprintf("http://%s", address)); err != nil {
+		errors.Wrap(err, "cannot open browser")
+	}
+
+	return nil
+}
+
+// OpenUI opens the browser with the UI.
+//
+// If opening the UI raises an error or took more time then `timeout`, then the
```
carlosms: multi-line suggestion:
```
// If opening the UI raises an error or took more time than `timeout`, then the
// error is sent to the returned channel.
```

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+
+func openUI() error {
+	var stdout, stderr bytes.Buffer
+	for {
+		if err := compose.RunWithIO(context.Background(),
+			os.Stdin, &stdout, &stderr, "port", "superset", "8088"); err == nil {
+			break
+		}
+
+		time.Sleep(1 * time.Second)
```
carlosms: I'm not sure we want to have this loop retrying.
If the containers are stopped, returning here with a message along the lines of "could not get superset port, is it running?" would be more clear than returning after 10s with "opening the ui took more than"

smacker: it makes sense to keep loop for `install` command.
depends on the hardware and other loads on the host system it can take a while.

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	return nil
+}
+
+// OpenUI opens the browser with the UI.
+//
+// If opening the UI raises an error or took more time then `timeout`, then the
+// error sent to the returned channel.
+func OpenUI(timeout time.Duration) <-chan error {
```
smacker: no need to return channel. For all our use cases it's absolutely okay to block in this function and return error or nil when it finishes.

smacker: looks like you needed it because `install` was running without `--detach`. now it does.

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+
+	return nil
+}
+
+func init() {
+	rootCmd.AddCommand(&webCmd{})
+}
+
+func openUI() error {
+	var stdout, stderr bytes.Buffer
```
smacker: you don't use `stderr` anywhere. Most probably you can just pass `nil` instead.

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,86 @@
...
+		time.Sleep(1 * time.Second)
+	}
+
+	address := strings.TrimSpace(stdout.String())
+	if address == "" {
+		return fmt.Errorf("no address found")
+	}
+
+	for {
+		if _, err := net.Dial("tcp", address); err == nil {
```
smacker: this would check only that the port is open (and it is open for sure because docker-compose opened it and there is a check already above)
But it doesn't mean anything listens on it yet. You need to use http.Client or user would get "The connection was reset" or something like that.

smacker:
I couldn't push into this branch so I created new PR: https://github.com/src-d/superset-compose/pull/49

PR #33, opened 2019-05-08
Add a UAST modal to SqlLab tab

carlosms:
Contains the same frontend code as in https://github.com/src-d/superset/pull/1.

To upgrade requirements.txt I had to upgrade `pip-tools` to 3.6.1. The version shipped in `requirements.txt` was not working for me, `pip-compile --output-file requirements.txt setup.py` failed.
The new `pip-tools` was upgrading the version for `flash-appbuilder`, so I pinned it to the previously installed version.

The backend code in this PR is different from https://github.com/src-d/superset/pull/1. In that one the json encoder was hijacked to unmarshal bytes as UAST JSON string. But that json encoder is not consistently used in the Superset code. For example an async call will try to store a query result as JSON as an intermediate step, and that encoding will fail is there is binary data that is not UTF8.

With the code in this PR the data is intercepted as it is received from the sql driver, and if a UAST binary column is found it gets replaced with a string containing the JSON of that UAST.

image

image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -53,7 +53,8 @@ RUN pip install --upgrade setuptools pip \
 COPY --chown=superset:superset superset/superset superset
 
 ENV PATH=/home/superset/superset/bin:$PATH \
-    PYTHONPATH=/home/superset/superset/:$PYTHONPATH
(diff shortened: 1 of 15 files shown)


carlosms:
rebased

PR #36, opened 2019-05-10
Add basic travis conf for releases

carlosms:
This PR adds basic travis conf to build the docker image on tag.
One downside of the configuration is that we will need to tag and release both the go cmd and the docker image, even if we only want to release one.
But I think this is acceptable for now, there is not need to create complex rules to build one or the other depending on the tag name. At some point we may want to divide the project and have the go cli live in a separate repo.

The docker-compose file for now needs `smacker/superset:demo-with-bblfsh`, so in order to build and test locally we need to run it as `VERSION=demo-with-bblfsh make superset-build`. When we do the first proper release under the `src-d` org I guess we'll want to change the version to be latest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,48 @@
+branches:
+  only:
+    - master
+    - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
+
+dist: xenial
+sudo: required
+
+language: go
+go_import_path: github.com/src-d/superset-compose
+go:
+  - 1.12.x
+env:
+  - GO111MODULE=on
+
+matrix:
+  fast_finish: true
+
+services:
+  - docker
+
+stages:
+  - name: tests
+  - name: release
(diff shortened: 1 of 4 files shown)


carlosms:
After the repo rename I have force-pushed with the docker image name `src-d/superset`.
I've put `latest` tag in the docker-compose file to make it easier to release for now. In the future we may probably want to pin it to some version.

Because the docker-compose file uses `latest`, to test locally the project needs to be built with `VERSION=latest make superset-build`

smacker:
CI is failing though

carlosms:
build was failing because `go_import_path` was set to `github.com/src-d/superset-compose`, but the code was using the older repo name. Fixed with #41. And then it couldn't find the dependencies, fixed setting the env var `GO111MODULE=on`.

Already rebased and squashed, travis is green.

PR #37, opened 2019-05-13
Initial commit of source{d} Theme V1

ricardobaeta:
Signed-off-by: Ricardo Baeta <ricardo@ricardobaeta.com>

image


ricardobaeta:
@maxim I agree and it makes sense for us to have a separate file. It's a good idea to have a standard way to customize the UI, that I find it would be a good contribution to the community. The only question that arises is the name of the file. If we want to contribute, the file could be named "custom-brand.less"?

ricardobaeta:
@smacker Done!

PR #49, opened 2019-05-16
Web ui opening

smacker:
fix: #17 replaces #31

1. rebased on master
2. applied review comments
3. changed the timeouts (as explained in commit message 10s isn't enough on my system for `start` and 10s is way too much for `web`)

### DIFF
--- a/cmd/sandbox-ce/cmd/web.go
+++ b/cmd/sandbox-ce/cmd/web.go
@@ -4,14 +4,14 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"net"
+	"net/http"
 	"os"
 	"strings"
 	"time"
 
 	"github.com/pkg/browser"
 	"github.com/pkg/errors"
-	"github.com/smacker/superset-compose/cmd/sandbox-ce/compose"
+	"github.com/src-d/superset-compose/cmd/sandbox-ce/compose"
 )
 
 type webCmd struct {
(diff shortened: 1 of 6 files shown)


cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,79 @@
...
+	"strings"
+	"time"
+
+	"github.com/pkg/browser"
+	"github.com/pkg/errors"
+	"github.com/src-d/superset-compose/cmd/sandbox-ce/compose"
+)
+
+type webCmd struct {
+	Command `name:"web" short-description:"Web"`
```
carlosms: Let's add both short and long descriptions. Something like "Open the web interface on your browser"

smacker:
descriptions added and rebased on master

cmd/sandbox-ce/cmd/web.go
```diff
@@ -0,0 +1,79 @@
...
+	ch := make(chan error)
+	go func() {
+		ch <- openUI()
+	}()
+
+	select {
+	case err := <-ch:
+		return errors.Wrap(err, "an error occurred while opening the UI")
+	case <-time.After(timeout):
+		return fmt.Errorf("opening the UI took more than %v", timeout)
```
carlosms: Small nitpick, but if I run `web` with the containers stopped and it says `opening the UI took more than 2s` it can be unclear if this is an error or a statement of how much the successful opening took.
```suggestion
		return fmt.Errorf("error opening the UI, the container is not running after %v", timeout)
```

Issue #14, opened 2019-05-02
sandbox-ce sub command to see components status


carlosms:
From the initial design doc:
> As a user I want to see all the components I have installed so that I know their version and their current status

At first sight I looks to me like this could be covered by a call to `docker-compose images` & `docker-compose ps`.

smacker:
it conflicts with requirements from Maximo here: https://docs.google.com/document/d/1l_q3xDs9FwY_oeGOhIXdV5R_0uFGd_rgOR23fO-d1CU/edit

se7entyse7en:
@smacker could you quote the part from the design doc?

carlosms:
friendly ping @smacker. What are the conflicts?

smacker:
This command is clone of `components list` of engine. And in the document it was rejected by Maximo:

<img width="577" alt="Screenshot 2019-05-16 at 11 28 57" src="https://user-images.githubusercontent.com/406916/57843012-d8c06800-77cd-11e9-96a0-bd525d8554f8.png">

Personally I also don't see much sense in such command.
If any of the compoments is down UI will be broken fully or partually. Which means everything should be running always or everything should be stop.

About the versions I also don't see much value in it. We are going to provide docker-compose.yml file with versions of the components that are tested and work together well. If the user want to change them it's advanced user use case. Though we still cover this use case by allowing the user to see and edit docker-compose.yml. It has all the versions there.

If we would want to give more flexibility to the user without low level knowleadge it's better to create new page in UI and combine it with bblfsh driver versions also than adding more cli commands.

mcuadros:
`sandbox-ce status` === `docker ps`

PR #50, opened 2019-05-16
Add status command

smacker:
fix: #14

### DIFF
--- a/cmd/sandbox-ce/cmd/status.go
+++ b/cmd/sandbox-ce/cmd/status.go
@@ -7,7 +7,7 @@ import (
 )
 
 type statusCmd struct {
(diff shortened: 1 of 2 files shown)


cmd/sandbox-ce/cmd/status.go
```diff
@@ -0,0 +1,19 @@
+package cmd
+
+import (
+	"context"
+
+	"github.com/src-d/superset-compose/cmd/sandbox-ce/compose"
+)
+
+type statusCmd struct {
+	Command `name:"status" short-description:"Shows status of the components"`
```
carlosms: Let's add `long-description` too, even if it has duplicated text.

smacker:
rebased on master & added long-description.

PR #51, opened 2019-05-16
Download docker-compose.yml if it doesn't exist

smacker:
It doesn't implement "download" command and doesn't handle update of the
config file. It only checks if the config exists and if it doesn't
downloads it from master branch. So the cli can be usable. Currently if
a one downloads it and try to run there is an error:

```
ERROR:
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml

exit status 1
```

### DIFF
--- a/cmd/sandbox-ce/compose/compose.go
+++ b/cmd/sandbox-ce/compose/compose.go
@@ -18,11 +18,15 @@ import (
 // composeContainerPath is the url docker-compose is downloaded from in case
 // that it's not already present in the system
 const composeContainerPath = "https://github.com/docker/compose/releases/download/1.24.0/run.sh"
+const configURL = "https://raw.githubusercontent.com/src-d/superset-compose/master/docker-compose.yml"
 
 var envKeys = []string{"GITBASE_REPOS_DIR"}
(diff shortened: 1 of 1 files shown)


carlosms:
I'm not too sure we should add this download without having a way to update the yml file. If we merge this wihtout #29, we would need to add docs to explain that users need to remove their `~/.srcd/docker-compose.yml` when they download a new release. Maybe we should wait and manage everything together in #29.

That's why I created #57, to be able to have working releases without having to rush other options.

smacker:
I would disagree. I think it's better to do it step by step.
- This PR enables downloading from master
- Next PR changes master to the last stable release
- Another PR allows to update yml file.

If we put everything in one PR it will be too big.

carlosms:
I agree smaller PRs makes development and review easier. My concern was about having master ready to tag a release at any time, I was thinking about doing this step by step in a feature branch that we could merge later to master.

But at this stage of the project, I guess it's ok if we don't want to be too strict.

smacker:
rebased on master. This PR introduces hard-coded project name but it's hard-coded only in the constructor and can be replaced by env variable value later.

Issue #35, opened 2019-05-09
Stopping queries running with Gitbase doesn't work


se7entyse7en:

carlosms:
More context: queries seem to be cancelled correctly from superset point of view, there are no errors reported to the user. But the query will continue running in gitbase, which means superset cancels by forgetting about the results, but not calling `KILL`.

smacker:
Superset doesn't really stop query.
The source code of the stop endpoint: https://github.com/src-d/superset-compose/blob/master/superset/superset/views/core.py#L2456
It only updates Query model instance in database changing the status.

The model: https://github.com/src-d/superset-compose/blob/master/superset/superset/models/sql_lab.py#L36
It doesn't have any hook that would cancel underlying queries.

Running query calls `db_engine_spec.execute` at the end: https://github.com/src-d/superset-compose/blob/master/superset/superset/sql_lab.py#L191

There is no hook to check status and cancel the query on EngineSpec level. `handle_cursor` is called only when the query is finished.

I'm currently exploring how difficult it is to add query cancelation support.

smacker:
I implemented cancellation: https://github.com/src-d/superset-compose/pull/53/files
But it doesn't work due to a bug in gitbase: https://github.com/src-d/gitbase/issues/817

The PR also doesn't have tests. Most probably after tests are written it also makes sense to rebase on superset master and suggest the change to upstream.

smacker:
fixed

PR #53, opened 2019-05-17
[sqllab] Cancel query to database on query stop

smacker:
- Add new attribute into Query model that keeps ID of an underlying
connection to database for this query
- Add `get_connection_id` method to EngineSpec that retrieves connection
id for a cursor.
- `execute_sql_statements` before executing queries retrieves and saves
connection id into Query
- Stop handler calls new `cancel_query` method of EngineSpec

Both new methods are implemented only for MySQL for now.
`cancel_query` uses KILL CONNECTION to terminate the statement.

Ref: #35

### DIFF
--- a/superset/superset/sql_lab.py
+++ b/superset/superset/sql_lab.py
@@ -68,6 +68,7 @@ def handle_query_error(msg, query, session, payload=None):
     query.error_message = msg
     query.status = QueryStatus.FAILED
     query.tmp_table_name = None
+    query.connection_id = None
     session.commit()
     payload.update({
         'status': query.status,
@@ -248,6 +249,8 @@ def execute_sql_statements(
     # execution of all statements (if many)
     with closing(engine.raw_connection()) as conn:
         with closing(conn.cursor()) as cursor:
+            query.connection_id = db_engine_spec.get_connection_id(cursor)
+
             statement_count = len(statements)
             for i, statement in enumerate(statements):
                 # TODO CHECK IF STOPPED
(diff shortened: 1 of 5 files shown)


smacker:
Tested with gitbase rc3 and it works like a charm.
It's not clear to me where and how to write tests for it, looks like superset test coverage is rather poor.
I'll propose this improvement to the upsteam and also ask for advice about tests.
But taking into account how important this feature is for our use cases I think this PR is good to be merged after code review.

superset/superset/sql_lab.py
```diff
@@ -193,7 +194,8 @@ def execute_sql_statement(sql_statement, query, user_name, session, cursor):
             db_engine_spec.handle_cursor(cursor, query, session)
 
         with stats_timing('sqllab.query.time_fetching_results', stats_logger):
-            logging.debug('Fetching data for query object: {}'.format(query.to_dict()))
+            logging.debug(
+                'Fetching data for query object: {}'.format(query.to_dict()))
```
carlosms: here

superset/superset/sql_lab.py
```diff
@@ -308,9 +313,28 @@ def execute_sql_statements(
             cache_timeout = database.cache_timeout
             if cache_timeout is None:
                 cache_timeout = config.get('CACHE_DEFAULT_TIMEOUT', 0)
-            results_backend.set(key, zlib_compress(json_payload), cache_timeout)
+            results_backend.set(key, zlib_compress(
+                json_payload), cache_timeout)
```
carlosms: here

smacker:
thanks! fixed & rebased on master

Issue #47, opened 2019-05-16
Proxy bblfsh-web api through superset


smacker:
Currently, UAST Lab connects directly to bblfsh-web server which is alright for local deployment.

But if sandbox-ce is deployed somewhere most probably it will expose only 1 public port of superset.

So we need to:
- proxy `/bblfsh/api` to bblfsh-web
- update configuration for bblfsh-web ui to use address from the browser instead of hard-coded `0.0.0.0:9999`

PR #55, opened 2019-05-20
add proxy for bblfsh-web server

smacker:
fix: #47


smacker:
Better not `/` because superset itself can be behind proxy with non-root path.

I also don't see problems with this generic name. Most of endpoints exposed by `bblfsh/api` are rather generic too. Like parse, drivers. If we need another endpoint we can always add it by exposing another view with `@expose('/bblfsh/api/whatever')`

PR #57, opened 2019-05-20
Add docker-compose.yml to release assets

carlosms:
The binary included in the 0.0.1 needs to have the docker-compose file downloaded manually, but this is not documented.
The important part of this PR is the changes in the README. This is just a way to make the 0.0.1 release work

With the changes in this PR the docker-compose.yml file will be uploaded to the release assets. If this one is approved I'll update the 0.0.1 release manually with the compose file as it was in that tag.
This file upload on travis.conf is not an alternative to #13, I added just in case we want to release again before #13 is done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,9 @@ jobs:
         provider: releases
         api_key: $GITHUB_TOKEN
         file_glob: true
-        file: build/*.tar.gz
+        file:
+          - build/*.tar.gz
+          - docker-compose.yml
         skip_cleanup: true
         on:
           all_branches: true
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ You may also choose to manage the containers yourself with the `docker-compose.y
 
 ### Commands
 
+Go to the [releases page](https://github.com/src-d/superset-compose/releases) and download the `sandbox-ce` binary for your system. You will also need to download the `docker-compose.yml` file included in the release assets.
+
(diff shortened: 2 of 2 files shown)


smacker:
taking into account that cli & docker-compose.yml will follow separate release cycles *soon* I would prefer using this https://github.com/src-d/superset-compose/pull/51 solution from the start and improve it later.

carlosms:
The start is 0.0.1, which is already released, and can be used without any problems _if_ we document that you need to download the compose file.

This doesn't have to be the way to do things from now on, it's just supposed to fix what is broken in 0.0.1, which is mainly a documentation problem.

smacker:
> The start is 0.0.1, which is already released

there is no problem with removing it and releasing it again with newer code.

carlosms:
I don't understand your concerns on why we should not merge this.
We agreed to release on the 16th, and the only problem with 0.0.1 is that it is missing documentation about the need of the docker-compose.yml file.

We can release again a 0.0.2 that handles the compose file in a better way, but after this PR we won't be pressured to release it right now.

smacker:
My concern is to fix v0.0.1 right now we can just update release notes mentioning the need to download docker-compose.yml manually.

You are fixing only next releases with this PR. And in my opinion for the next release it's better to implement something on top of Marvin's and my PRs than proposed change. Though it doesn't hurt to merge. If you really want to - go ahead.

PR #59, opened 2019-05-20
Move docker files to superset sub dir

### LINKED ISSUE #54: Move docker dir into superset subtree
I didn't do it when added superset into the repository.

carlosms:
Fix #54.

The build now uses the `superset/contrib/docker` dir for the docker files, this way we can keep up with changes in upstream (there were a few small changes already that were missing from our Dockerfile).

After moving the docker related files I've realized that the `.env` file was setting the docker compose project name, but this is not very reliable. If you download the binary from our releases page you will not have this file. So I added the `--project-name` option to the `docker-compose` command. I also tested this for the compose inside a container.


superset/contrib/docker/.env
```diff
@@ -1,17 +0,0 @@
-#
```
smacker: why do you need to delete it?

cmd/sandbox-ce/compose/compose.go
```diff
@@ -135,5 +135,6 @@ func Run(ctx context.Context, arg ...string) error {
 		return err
 	}
 
-	return comp.Run(ctx, arg...)
+	newArgs := append([]string{"--project-name", "srcd"}, arg...)
```
smacker: github doesn't show conflict but current master already set project name

superset/contrib/docker/superset_config.py
```diff
@@ -1,58 +0,0 @@
-# Licensed to the Apache Software Foundation (ASF) under one
```
smacker: maybe we should move our config here too?

superset/contrib/docker/.env
carlosms: We don't rely on the contents of `.env` for a normal installation of `sandbox-ce`, so I wanted to notify this somehow here. If you add variables here, they will not have an effect on our final distribution.

superset/contrib/docker/superset_config.py
carlosms: I was going to say no because right now `superset_config` and `superset_config_dev` are in the same dir. But moving it here allows us to detect if there's anything added in upstream, so makes sense. Done in df208b52.

cmd/sandbox-ce/compose/compose.go
carlosms: in master it was a different method, so there was no conflict but the result would be the same arg added twice. I have removed it here, rebased on top of master and squashed.

PR #62, opened 2019-05-21
Make TEXT columns filterable and groupable

### LINKED ISSUE #34: Tables' columns are nor groupable nor filterable by default
A column can be `groupable`, `filterable` and `is_temporal`.

The columns of the imported gitbase tables are nor `groupable` nor `filterable` by default. So before creating a chart that needs grouping or filtering on some columns, the user needs to get to the table configuration and mark the columns as manually.

Example of some columns manually marked both as `groupable` and `filterable`.

Screenshot 2019-05-09 at 11 45 59

Marking them during the creation of the tables would improve the user experience.

On the other hand, each column with type `TIMESTAMP` is already automatically marked as `is_temporal`.

carlosms:
Fix #34.

`fetch_metadata` already [makes string columns filterable and groupable](https://github.com/src-d/superset-compose/blob/master/superset/superset/connectors/sqla/models.py#L879), but TEXT was not included as a string type.
We might want to PR this to upstream also.


smacker:
Hm. `TEXT` type is meant to be huge text that doesn't make much sense to group of filter by user.
I mean maybe it's okay to fix as proposed right now. But better fix would be to return correct types from gitbase. Then for example repository name would be checked automatically but commit message not.

smacker:
Here is the issue for gitbase: https://github.com/src-d/gitbase/issues/824
It doesn't look like this issue is very critical. I would better wait for fix in gitbase than hacking superset.

PR #63, opened 2019-05-21
Add restart policy to gitbase container

### LINKED ISSUE #61: Add restart policy on gitbase
It might be worth to add a restart policy in docker-compose for gitbase, if it crashes, we can restart it (with a maximum number of attempts within a timeframe).

See the restart policy section on the [docker compose reference](https://docs.docker.com/compose/compose-file/).

carlosms:
Fix #61.

According to docs, [`deploy/restart_policy`](https://docs.docker.com/compose/compose-file/#restart_policy) only takes effect when deploying to a swarm with `docker stack deploy`, and is ignored by `docker-compose up`.

The closest thing we have is [`restart`](https://docs.docker.com/compose/compose-file/#restart).


carlosms:
:+1: I have added restart to bblfsh. Rebased and squashed.

friendly ping @smola for review.

PR #48, opened 2019-05-16
create default user

smacker:
fix: #11

PR #66, opened 2019-05-23
Store compose files in workdirs inside ~/.srcd

carlosms:
Part of #15.

This is my proposal to handle multiple deployments for different sets of repositories.

This should give you a good picture of the idea:
```
$ sandbox-ce install ~/go/src/github.com/src-d/
$ sandbox-ce install ~/repos

$ tree -a ~/.srcd/
/home/cmartin/.srcd/
├── docker-compose.yml
└── workdirs
    ├── __active__ -> /home/cmartin/.srcd/workdirs/home/cmartin/repos
    └── home
        └── cmartin
            ├── go
            │   └── src
            │       └── github.com
            │           └── src-d
            │               ├── docker-compose.yml -> /home/cmartin/.srcd/docker-compose.yml
            │               └── .env
            └── repos
                ├── docker-compose.yml -> /home/cmartin/.srcd/docker-compose.yml
                └── .env

$ pwd
/home/cmartin/.srcd/workdirs/__active__

$ cat .env 
GITBASE_REPOS_DIR=/home/cmartin/repos
COMPOSE_PROJECT_NAME=srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b

$ docker-compose ps
                           Name                                         Command                  State               Ports         
-----------------------------------------------------------------------------------------------------------------------------------
srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_bblfsh-web_1   /bin/bblfsh-web -addr :808 ...   Up             0.0.0.0:9999->8080/tcp
srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_bblfsh_1       bblfshd                          Up             0.0.0.0:9432->9432/tcp
srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_gitbase_1      ./init.sh                        Up             0.0.0.0:3306->3306/tcp
srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_postgres_1     docker-entrypoint.sh postgres    Up             0.0.0.0:5432->5432/tcp
srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_redis_1        docker-entrypoint.sh redis ...   Up             0.0.0.0:6379->6379/tcp
srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_superset_1     /entrypoint.sh                   Up (healthy)   0.0.0.0:8088->8088/tcp

$ docker volume ls
DRIVER              VOLUME NAME
local               srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_postgres
local               srcd-4c901c8fd0b755feb1470a91dc59d50d806c748b_redis
local               srcd-ef2f6172f3165cd61fdcfc765ac9440610d2dbb5_postgres
local               srcd-ef2f6172f3165cd61fdcfc765ac9440610d2dbb5_redis
```

Key docker compose behaviours leveraged in this PR:
- Docker compose already handles different deployments isolation with the `project name`
- This project name by default is the docker-compose.yml dir name. It can be also set with COMPOSE_PROJECT_NAME
- docker-compose will read env vars from a `.env` file. This file needs to be in the current working directory. This means `docker-compose -f /tmp/docker-compose.yml` will ignore `/tmp/.env` unless your current dir is `/tmp`.

The isolation happens because each dir in `.srcd/workdirs` contains a unique COMPOSE_PROJECT_NAME made with the full path sha1.

The sha1 makes the docker names ugly, and hard to relate to some specific folder. But the full name must be used to create the project name, otherwise `~/repos` and `~/other/path/repos` would share the `repos` project name. Another alternative might be to use the full path in clear, but something like `srcd-home_cmartin_repos_postgres` is also a bit awkward to manage, as it can grow to a very long name.

Each dir could be placed as a first level sub dir of `.srcd/workdirs`, using `.srcd/workdirs/<sha1>`. But I think having the full folder hierarchy replicated in `.srcd/workdirs`  makes it intuitive for advanced users to just `cd ~/.srcd/workdirs/` and understand what is going on.

From any of the directories there you can run any `docker-compose` commands without any extra requirements. No env vars to set, no extra options like `--file`.

All the `docker-compose.yml` files are symbolic links. So when the mechanism to download newer compose files is in place we will not have to update anything in each workdir.

If you try it out you'll see that `install` will prompt you to create the admin user even when you use a dir that was already initialized. This will be solved in #48.

I didn't write any docs or changelog because I wanted your feedback on it first.

Also there are missing details that I think are better addressed in future PR, and also make sense to include in future releases.
- Handle empty `.srcd` when other commands are executed before `install`
```
$ rm -rf ~/.srcd
$ sandbox-ce stop
chdir /home/cmartin/.srcd/workdirs/__active__: no such file or directory
```
- List all the existing workdirs. This can be also useful to `prune` everything done by `sandbox-ce`, not just the current `__active__` volumes.
- Consider a rename of the command. `install` is fine for the first time you run the sandobx. But as a command to come back to a dir that was used previously, maybe `init` would be more intuitive. Or `up`, to make it clear that it's the equivalent of `docker-compose up`

### DIFF
--- a/cmd/sandbox-ce/compose/compose.go
+++ b/cmd/sandbox-ce/compose/compose.go
@@ -2,20 +2,16 @@ package compose
 
 import (
 	"context"
-	"crypto/sha1"
-	"encoding/hex"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
 
-	composefile "github.com/src-d/superset-compose/cmd/sandbox-ce/compose/file"
+	"github.com/src-d/superset-compose/cmd/sandbox-ce/compose/workdir"
 	"github.com/src-d/superset-compose/cmd/sandbox-ce/dir"
-	datadir "github.com/src-d/superset-compose/cmd/sandbox-ce/dir"
 
 	"github.com/pkg/errors"
 )
@@ -25,8 +21,6 @@ import (
 const composeContainerURL = "https://github.com/docker/compose/releases/download/1.24.0/run.sh"
 
 var envKeys = []string{"GITBASE_REPOS_DIR"}
-
-const activeWorkdir = "__active__"
 
 type Compose struct {
 	bin string
@@ -40,7 +34,7 @@ func (c *Compose) RunWithIO(ctx context.Context, stdin io.Reader,
 	stdout, stderr io.Writer, arg ...string) error {
 	cmd := exec.CommandContext(ctx, c.bin, arg...)
 
-	dir, err := workdirPath(activeWorkdir)
+	dir, err := workdir.Active()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


cmd/sandbox-ce/compose/compose.go
```diff
@@ -204,3 +218,104 @@ func expandHome(path string) (string, error) {
...
+
+	workdir := filepath.Join(homedir, ".srcd", "workdirs", reposdir)
+
+	err = os.MkdirAll(workdir, 0755)
+	if err != nil {
+		return "", errors.Wrap(err, "could not create working directory")
+	}
+
+	composePath := filepath.Join(workdir, "docker-compose.yml")
+	_, err = os.Stat(composePath)
```
smacker: need a check for `err != nil && !os.IsNotExist(err)`

carlosms: good catch, done in 5667c969

PR #67, opened 2019-05-24
Store gitbase indexes in a volume

### LINKED ISSUE #65: Add persistence to gitbase index data
I haven't tested this, but I think with the current config the gitbase container will loose the index data after a `docker-compose down`. If it is stored in a volume this data will persist after a `down; up`

carlosms:
Fix #65

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,7 @@ services:
         target: /opt/repos
         read_only: true
         consistency: delegated
+      - gitbase_indexes:/var/lib/gitbase/index
 
   bblfsh-web:
     image: bblfsh/web:v0.11.0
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
         target: /opt/repos
         read_only: true
         consistency: delegated
+      - indexes:/var/lib/gitbase/index
```
smacker: 1. space
2. let's call it `gitbase_indexes` or something like that. `indexes` is way too general name.

carlosms: > space

what do you mean?

smacker: ah. My bad. It is just below lines with syntax highlighted but this line not. But actually it's correct.

carlosms: :+1: name changed

Issue #77, opened 2019-05-29
Allow to configure gitbase spark connector via environmental variables


rporres:
I need to be able choose between gitbase or gitbase spark connector via configuration variables rather than adding a datasource from superset. That also means that gitbase configuration via env variables should not be mandatory or at least check if either gitbase or gitbase spark connector configuration is present.

smacker:
taking into account that we will use spark-connector only for EE and gitbase only for CE, I would propose:
1. new env variable `MODE: Community|Enterprise`
2. depends on it we will use different prefixes for datasource mysql/hive but the same current variables to build the connection string: GITBASE_HOST/PORT/...
3. later we can use the same variable to load different dashboards and do whatever other changes we need.

smacker:
Env var is merged in master.
But our superset can't work with spark yet because it requires patched PyHive so I keep the issue open for now.
We have a PR in our PyHive fork but it's not merged yet: https://github.com/src-d/PyHive/pull/1

smacker:
fixed by https://github.com/src-d/sourced-ui/pull/87

PR #79, opened 2019-05-30
use sparksql instead of mysql for enterprise version

smacker:
New optional environmental variable MODE set into Enterprise switches
connection string prefix to sparksql.

I didn't added variable into docker-compose and also didn't add docs because it's not clear for me if we should document EE publicly or not.

### DIFF
--- a/superset/contrib/docker/superset_config.py
+++ b/superset/contrib/docker/superset_config.py
@@ -70,16 +70,19 @@ CELERY_CONFIG = CeleryConfig
 
 # Gitbase configuration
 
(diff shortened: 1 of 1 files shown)


carlosms:
Fix #77?

smacker:
It partly fixes #77. I mean it implements what is requested but our superset can't work with spark yet because it requires patched PyHive.
But remember that it's not going to work yet because we need patched PyHive.
We have a PR in our PyHive fork but it's not merged yet: https://github.com/src-d/PyHive/pull/1

Issue #78, opened 2019-05-29
Split this repository into sourced-ui and sourced-ce


smacker:
### sourced-ce

Move cli, docker-compose.yml, docs, ...:
```
git co -b sourced-ce
git filter-branch --index-filter 'git rm --cached -qr --ignore-unmatch superset' sourced-ce
git filter-branch --index-filter 'git rm --cached -qr --ignore-unmatch srcd' sourced-ce
git filter-branch -f --prune-empty --parent-filter 'sed "s/-p //g" | xargs git show-branch --independent | sed "s/^/-p /g" | tr "\n" " "' -- sourced-ce
```

### sourced-ui
Instructions how to move superset into separate repo (and make changes that are required to use `git subtree` instead of low-level git commands):
https://gist.github.com/smacker/f0a317d5e618eedbc5b77ba36cad1f0c

Instructions how to merge new release using `git subtree merge`. (in case of hard-fork there will be not much differences as long as we want to maintain our master based on different releases):
https://gist.github.com/smacker/b8a6ccf5539ad2a209c97fe878233ac7

PR #85, opened 2019-05-31
Copy back missing docs and build scripts

carlosms:
Part of #78.

Instead of overriding `build` I left `superset-build` as a new target. Without `COMMANDS` defined the go `build` does nothing, and this way `make build` does not complain about ignored commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,16 @@
 # Package configuration
-PROJECT = sandbox-ce
+PROJECT = sourced-ui
 
 # superset upstream configuration
 SUPERSET_REPO = https://github.com/apache/incubator-superset.git
-SUPERSET_VERSION = release--0.32
+SUPERSET_VERSION = 0.32.0rc2
 SUPERSET_REMOTE = superset
 # directory to sync superset upstream with
 SUPERSET_DIR = superset
 # directory with custom code to copy into SUPERSET_DIR
 PATCH_SOURCE_DIR = srcd
 # name of the superset docker image to build
-SUPERSET_IMAGE_NAME = srcd/superset
+SUPERSET_IMAGE_NAME = srcd/sourced-ui
 
 # Including ci Makefile
 CI_REPOSITORY ?= https://github.com/src-d/ci.git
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,94 @@
...
+
+# Prints unified diff of local superset  and upstream
+.PHONY: diff
+diff: superset-remote-add
+	git diff-tree -p $(SUPERSET_REMOTE)/$(SUPERSET_VERSION) HEAD:$(SUPERSET_DIR)/
+
+# Merge remote superset into SUPERSET_DIR as squashed commit
+.PHONY: merge
+merge: superset-remote-add
+	git merge --squash -s subtree --no-commit remotes/$(SUPERSET_REMOTE)/$(SUPERSET_VERSION)
```
smacker: could you please remove this? we are using `git subtree` now.

Makefile
```diff
@@ -0,0 +1,94 @@
...
+# superset upstream configuration
+SUPERSET_REPO = https://github.com/apache/incubator-superset.git
+SUPERSET_VERSION = release--0.32
+SUPERSET_REMOTE = superset
+# directory to sync superset upstream with
+SUPERSET_DIR = superset
+# directory with custom code to copy into SUPERSET_DIR
+PATCH_SOURCE_DIR = srcd
+# name of the superset docker image to build
+SUPERSET_IMAGE_NAME = srcd/superset
```
smacker: maybe it should be `srcd/sourced-ui` (I don't know)

Makefile
```diff
@@ -0,0 +1,94 @@
+# Package configuration
+PROJECT = sandbox-ce
```
smacker: sourced-ui

Makefile
```diff
@@ -0,0 +1,94 @@
...
+
+# superset upstream configuration
+SUPERSET_REPO = https://github.com/apache/incubator-superset.git
+SUPERSET_VERSION = release--0.32
```
smacker: `0.32.0rc2`

Makefile
carlosms: do `diff` and `diff-stat` need any change?

README.md
```diff
@@ -1 +1,58 @@
...
+
+## Contents
+
+- [Development](#development)
+
+## Development
+
+### Setup local environment
+
+Run dependencies using docker-compose:
```
smacker: hm... but the repository doesn't have docker-compose.yml anymore

Makefile
carlosms: yes, keeping the same name as the repo sounds better.

Makefile
carlosms: this was changed during the project rename? Because I copied the values from `superset-compose` master.

Makefile
smacker: nope. There are no commands for diff in `git subtree`.

Makefile
smacker: yep: https://gist.github.com/smacker/f0a317d5e618eedbc5b77ba36cad1f0c#file-move_superset-txt-L13
They actually point on the same commit most probably but still

carlosms:
I fixed all comments on 0f39b5ba.

Makefile
carlosms: I also created the PR to change it in the compose file https://github.com/src-d/sourced-ce/pull/5

smacker:
ok. I need it. Merging.

PR #87, opened 2019-05-31
Add gitbase-spark-connector support

smacker:
- Use patched version of PyHive
- Update config script to produce correct connection string without name
and password (currently gsc requires connection without user/pass)

### DIFF
--- a/superset/requirements-dev.txt
+++ b/superset/requirements-dev.txt
@@ -25,7 +25,7 @@ mysqlclient==1.3.13
 pip-tools==3.6.1
 psycopg2-binary==2.7.5
 pycodestyle==2.4.0
-pyhive==0.6.1
+-e git+https://github.com/src-d/PyHive.git@master#egg=PyHive
 pylint==1.9.2
 python-dotenv==0.10.1
 redis==2.10.6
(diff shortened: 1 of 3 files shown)


superset/requirements-dev.txt
```diff
@@ -25,7 +25,7 @@ mysqlclient==1.3.13
 pip-tools==3.6.1
 psycopg2-binary==2.7.5
 pycodestyle==2.4.0
-pyhive==0.6.1
+-e git+https://github.com/src-d/PyHive.git@master#egg=PyHive
```
carlosms: should we make a tag in our fork, instead of using master?

smacker: DP team is responsible for that repo. So I don't know. I work with what I have.

smola: cc @ajnavarro maybe we should tag something like `v0.6.1-srcd1`?

smacker:
Just to be clear: this PR allows to initialize source{d} with gsc as a datasource. It doesn't mean source{d} works with gsc. (I couldn't execute any query, only get a list of tables)
But we will improve it later. (most probably with some help from DP team)

smacker:
update: actually queries work. Only preview doesn't work because it selects from `schema.table` which doesn't work. Select from `table` works.

superset/requirements-dev.txt
ajnavarro: https://github.com/src-d/PyHive/releases/tag/v0.7.0-srcd1

Issue #89, opened 2019-06-03
Update to superset release 0.33


smacker:
Instruction is here: https://github.com/src-d/sourced-ui/issues/78
Then need manual testing (if I remember right css changed a bit. need amends to uast tab and maybe to popup in gitlab as well)

carlosms:
One thing to take into account is the DB upgrade.
Right now we skip `docker-init.sh` completely if the DB contains the admin user, but it also includes `superset db upgrade`
https://github.com/src-d/sourced-ui/blob/master/superset/contrib/docker/docker-entrypoint.sh#L20

carlosms:
Reopening to update the version in the README, and add it to the changelog.

carlosms:
Maybe for the readme let's just say it's based on superset, removing the exact version. And then in the release notes we will notify when we merge a new version, that should be enough.

smacker:
looks like it was done in https://github.com/src-d/sourced-ui/pull/101

PR #92, opened 2019-06-04
Update to superset release 0.33

smacker:
Fix: #89


smacker:
git subtree doesn't support `--signoff`. Rebasing would change commit hashes which can break something. I propose to just ignore DCO.

carlosms:
@smola is it ok to ignore the DCO in this case?

smola:
@carlosms We should probably sign-off everything.

PR #97, opened 2019-06-06
Add repo description to readme

carlosms:
Adds repo and image description as requested in the backlog issue.


carlosms:
I agree on the screenshot, but I'd prefer to have one based on the default dashboards that will be included eventually. This way they'll match you will find when you open it.
I'd prefer to wait for the dashboards and add screenshots here and in the sourced-ce readme.

Issue #95, opened 2019-06-05
Incorrect timeout for sync sql queries


smacker:
We had overridden timeout from default 60 seconds to 5 minutes as gunicorn parameter:
https://github.com/src-d/sourced-ui/blob/master/superset/contrib/docker/docker-entrypoint.sh#L36

But it doesn't work in v0.1.0 image (most probably superset version used there requires some other changes)

PR #98, opened 2019-06-06
Fix incorrect timeout value for charts

smacker:
Fix: #95


smacker:
Oh changelog. Re-pushed.

PR #99, opened 2019-06-07
Add SUPERSET_NO_DB_INIT

meyskens:
Setting this envvar to true disables the docker-init script.
This is required to set up a proper deploy inside Kubernetes.

Signed-off-by: Maartje Eyskens <maartje@eyskens.me>


carlosms:
Should we add this env var to the docs?

Maybe a new section inside Description:
```
## Description

blah blah

### Environment Variables

You can configure the docker image reads the following environment variables:

- blah blah
```

meyskens:
@smacker re-worked those changes a bit so I can just use the `docker-init.sh` in the Helm hooks

meyskens:
@carlosms sure! How about a table somehere like we did in https://github.com/src-d/terraform-provider-online#developing-the-provider?

carlosms:
A table would be perfect

meyskens:
@carlosms added a table with all envvars I knew about

superset/contrib/docker/docker-init.sh
```diff
@@ -17,23 +17,27 @@
...
-    --app superset \
-    --username $ADMIN_LOGIN \
-    --firstname $ADMIN_FIRST_NAME \
-    --lastname $ADMIN_LAST_NAME \
-    --email $ADMIN_EMAIL \
-    --password $ADMIN_PASSWORD
-
-# Initialize the database
+# always run migrations
 superset db upgrade
```
smacker: did you test it?
I see the order of `create-admin` and `db upgrade` are different. Maybe it doesn't actually matter but better check.

README.md
```diff
@@ -19,6 +19,33 @@ This repository contains the code for the [`srcd/sourced-ui`](https://hub.docker
 - Backport an upstream fix for Hive Database connection ([#21](https://github.com/src-d/sourced-ui/issues/21)).
 - Cancel database queries on stop ([#35](https://github.com/src-d/sourced-ui/issues/35)).
 
+
+### Environment Variables
```
smacker: thank you 🙇

superset/contrib/docker/docker-init.sh
meyskens: Tested it, works fine

Issue #93, opened 2019-06-05
The build system should not require Go


carlosms:
`build` and `clean` require `go` is installed, but it should not be a strict requirement to build.

smacker:
I would remove dependency on `srcd/ci` for this repo completely.
We need just to run `docker build` and `docker push`. It will not only solve this problem but simplify Makefile as well.

PR #100, opened 2019-06-07
Remove dependency on src-d/ci

smacker:
Fix: #93

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,39 +1,47 @@
-# Package configuration
-PROJECT = sourced-ui
-
-# superset upstream configuration
+# Superset upstream configuration
 SUPERSET_REPO = https://github.com/apache/incubator-superset.git
 SUPERSET_VERSION = 0.33.0rc1
 SUPERSET_REMOTE = superset
-# directory to sync superset upstream with
+# Directory to sync superset upstream with
 SUPERSET_DIR = superset
-# directory with custom code to copy into SUPERSET_DIR
+# Directory with custom code to copy into SUPERSET_DIR
 PATCH_SOURCE_DIR = srcd
-# name of the superset docker image to build
-SUPERSET_IMAGE_NAME ?= srcd/sourced-ui
 
-# Including ci Makefile
-CI_REPOSITORY ?= https://github.com/src-d/ci.git
-CI_PATH ?= $(shell pwd)/.ci
-CI_VERSION ?= v1
+# Name of the docker image to build
+DOCKER_IMAGE_NAME ?= srcd/sourced-ui
+# Docker registry where the docker image should be pushed to.
+DOCKER_REGISTRY ?= docker.io
+# Docker organization to be used at the docker image name.
+DOCKER_ORG ?= srcd
+# Username used to login on the docker registry.
+DOCKER_USERNAME ?=
+# Password used to login on the docker registry.
+DOCKER_PASSWORD ?=
+# When `make docker-push`, setting DOCKER_PUSH_LATEST to any non-empty value
+# will cause make docker-push to also push the latest tag.
+DOCKER_PUSH_LATEST ?=
 
-MAKEFILE := $(CI_PATH)/Makefile.main
-$(MAKEFILE):
-	git clone --quiet --branch $(CI_VERSION) --depth 1 $(CI_REPOSITORY) $(CI_PATH);
+# Build information
+VERSION ?= latest
+# Travis CI
+ifneq ($(TRAVIS_TAG), )
+	VERSION := $(TRAVIS_TAG)
+endif
 
--include $(MAKEFILE)
+# IS_RELEASE is "true" if tag is semantic version and not a pre-release
+IS_RELEASE := $(shell echo $(VERSION) | grep -q -E '^v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$$' && echo "true" || true)
 
-all: superset-remote-add
+all: build
 
 # Copy src-d files in the superset repository
 .PHONY: patch
-patch: superset-clean
+patch: clean
 	cp -r $(PATCH_SOURCE_DIR)/* $(SUPERSET_DIR)/
 
 # Copy src-d files in the superset repository using symlinks. it's useful for development.
 # Allows to run flask locally and work only inside superset directory.
 .PHONY: patch-dev
-patch-dev: superset-clean
+patch-dev: clean
 	@diff=`diff -r $(PATCH_SOURCE_DIR) $(SUPERSET_DIR) | grep "$(PATCH_SOURCE_DIR)" | awk '{gsub(/: /,"/");print $$3}'`; \
 	for file in $${diff}; do \
 		to=`echo $${file} | cut -d'/' -f2-`; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -42,48 +50,58 @@ patch-dev: superset-clean
...
 
-superset-docker-push-latest-release:
-	@DOCKER_PUSH_LATEST=$(IS_RELEASE) make superset-docker-push
+docker-push-latest-release:
+	@DOCKER_PUSH_LATEST=$(IS_RELEASE) make docker-push
 
 # Clean superset directory from copied files
-.PHONY: superset-clean
-superset-clean:
+.PHONY:
```
carlosms: should be this instead?
```suggestion
.PHONY: clean
```

smacker: thanks. fixed. Also fixed some other targets that didn't have `.PHONY`

Makefile
```diff
@@ -42,48 +50,62 @@ patch-dev: superset-clean
...
+		docker tag $(DOCKER_IMAGE_NAME):$(VERSION) \
+			$(DOCKER_IMAGE_NAME):latest; \
+		docker push $(DOCKER_IMAGE_NAME):latest; \
 	fi;
 
-superset-docker-push-latest-release:
-	@DOCKER_PUSH_LATEST=$(IS_RELEASE) make superset-docker-push
+.PHONY: docker-push-latest-release
+docker-push-latest-release:
+	@DOCKER_PUSH_LATEST=$(IS_RELEASE) make docker-push
```
dpordomingo: this is a copy-paste from [src-d/ci/Makefile::docker-push-latest-release](https://github.com/src-d/ci/blob/v1/Makefile.main#L221).
Why was it really needed before this PR? Maybe to differentiate sandbox and ui images? Not it is no longer needed.

Makefile
```diff
@@ -42,48 +50,62 @@ patch-dev: superset-clean
...
 
-# Push the superset docker image, based on .ci/Makefile.main
-superset-docker-push: docker-login superset-build
+.PHONY: docker-login
+docker-login: docker-validate
+	@docker login -u "$(DOCKER_USERNAME)" -p "$(DOCKER_PASSWORD)" $(DOCKER_REGISTRY); \
+
+# Push the docker image
+.PHONY: docker-push
+docker-push: docker-login build
```
dpordomingo: Isn't this a copy-paste from [src-d/ci/Makefile::docker-push](https://github.com/src-d/ci/blob/v1/Makefile.main#L204) that could be removed if using it?

Makefile
```diff
@@ -42,48 +50,62 @@ patch-dev: superset-clean
...
 # Create docker image
-.PHONY: superset-build
-superset-build: patch
-	docker build -t $(SUPERSET_IMAGE_NAME):$(VERSION) -f superset/contrib/docker/Dockerfile superset
+.PHONY: patch
+build: patch
+	docker build -t $(DOCKER_IMAGE_NAME):$(VERSION) -f superset/contrib/docker/Dockerfile $(SUPERSET_DIR)
+
+.PHONY: docker-validate
+docker-validate:
```
dpordomingo: Isn't this a copy-paste from [src-d/ci/Makefile::docker-validate](https://github.com/src-d/ci/blob/v1/Makefile.main#L183) that could be removed if using it?

Makefile
```diff
@@ -42,48 +50,62 @@ patch-dev: superset-clean
...
+	fi; \
+	if [ -z "$(DOCKER_PASSWORD)" ]; then \
+		echo "DOCKER_PASSWORD variable cannot be empty."; \
+		exit 1; \
+	fi
 
-# Push the superset docker image, based on .ci/Makefile.main
-superset-docker-push: docker-login superset-build
+.PHONY: docker-login
+docker-login: docker-validate
```
dpordomingo: Isn't this a copy-paste from [src-d/ci/Makefile::docker-login](https://github.com/src-d/ci/blob/v1/Makefile.main#L180) that could be removed if using it?

Makefile
```diff
@@ -1,39 +1,47 @@
...
+# Build information
+VERSION ?= latest
+# Travis CI
+ifneq ($(TRAVIS_TAG), )
+	VERSION := $(TRAVIS_TAG)
+endif
 
--include $(MAKEFILE)
+# IS_RELEASE is "true" if tag is semantic version and not a pre-release
+IS_RELEASE := $(shell echo $(VERSION) | grep -q -E '^v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$$' && echo "true" || true)
```
dpordomingo: Already in [src-d/ci/Makefile::IS_RELEASE](https://github.com/src-d/ci/blob/v1/Makefile.main#L68)

Makefile
```diff
@@ -1,39 +1,47 @@
...
+# will cause make docker-push to also push the latest tag.
+DOCKER_PUSH_LATEST ?=
 
-MAKEFILE := $(CI_PATH)/Makefile.main
-$(MAKEFILE):
-	git clone --quiet --branch $(CI_VERSION) --depth 1 $(CI_REPOSITORY) $(CI_PATH);
+# Build information
+VERSION ?= latest
+# Travis CI
+ifneq ($(TRAVIS_TAG), )
```
dpordomingo: Already in [src-d/ci/Makefile#L58](https://github.com/src-d/ci/blob/v1/Makefile.main#L58)

smacker:
Simplification doesn't always mean moving complexity somewhere else. It can improve clarity and simplicity for understanding, remove additional layers of abstraction.

CI repo was created and is used for building go projects. This repository isn't go project.
Overriding things would work of course but can easily break later. And it just doesn't feel right to me to use something complex that wasn't meant for this use case only to abstract 2 commands `docker build` & `docker push`.

dpordomingo:
I'll try to disregard the general conversation about complexity, simplicity, abstraction, and explicitness.

I don't see the problem of overriding unwanted targets, and reusing needed ones, even more, when those are so well defined:
- `buid` and `clean` are done in a different way,
- release, versioning, image handling and such are done in the same way.

The alternative is copy-pasting code from `a` to `b`, making it less maintainable, and also keeping the possibility of failing later.

But if the rest of the team also think we could proceed that way, let's do so; I just wanted to raise my concerns on this topic.

carlosms:
I don't have a strong preference because I agree with both problems.
We are duplicating code in here that is already in the `ci` repo, which is bad practice.
But bringing to our Makefile targets like `package` `test-race` and others that do not apply here at all is also weird.

If there was a way to include here only the targets for docker building that would be ideal, but I think that's not possible.
To me this PR is good enough.

dpordomingo:
From my pov, there is no problem on having some not used (but imported) make rules (it also happens in go projects, and nobody cared); but not trying to block the merge: go ahead with the merge, and let's improve `ci` in the future if needed :+1: 

thank you all for the conversation and for explaining your opinions!

smacker:
I agree it makes sense to improve ci repo that can be used here and in other projects like gsc, gemini, bblfsh/python-client and so on.
Maybe split Makefile into 2. One general purpose and one for Golang.

PR #101, opened 2019-06-07
Update changelog for v0.2.0

smacker:


carlosms:
Let's also remove the version from the readme, "This image is based on Apache Superset 0.32", leaving only "based on Apache Superset".

Issue #96, opened 2019-06-06
Gitbase datasource is not updated if we restart sourced-ui with a changed GITBASE_* env variable


rporres:
This makes impossible to reconfigure via a restart with env variable changed if something like GITBASE_PASSWORD has changed

smacker:
It's a little tricky from the distributed deployment point of view.
The datasource is saved in DB.
So most probably it should be updated only in DB migration job and should be disable when [SUPERSET_NO_DB_INIT](https://github.com/src-d/sourced-ui/pull/99) is passed. Is that correct?

meyskens:
Yup, we will run `docker-init.sh` on every update so you can safely place it in there

PR #104, opened 2019-06-11
Always run add_gitbase script

smacker:
So it would update datasource if env variables have changed.

Fix: #96


carlosms:
What happens if there are no changes? Do we get any error messages, or is just updated successfully wit the same values anyway?

smacker:
it updates successfully with the same value.

Issue #102, opened 2019-06-11
Create default dashboards


carlosms:
This issue is about bringing all the charts from the PoCs and demos that don't require any pre-preprocessing, and can be run with the current data provided by sourced-ce.

smacker:
list of charts in demo that use only gitbase:
- Repository Count
- Languages
- Commits Evolution
- Repositories by number of files
- Number of commits per day of the week (all projects)
- Compliance score
- Missing files
- Files missing summary

PR #105, opened 2019-06-11
Update welcome dashboard with real charts

smacker:
Only gitbase based charts:

- Repository Count
- Languages
- Commits Evolution
- Repositories by number of files
- Number of commits per day of the week
- Compliance score
- Missing files
- Files missing summary

All of them transfered manually from demo.

I put all of them on the same dashboard without splitting into tabs
because there are just few of them. We can re-arrange when we have more.

Fix: #102

Screenshot from dataset `~/go/src/github.com/src-d/`
Screenshot_2019-06-11 Welcome(1)

Screenshot from dataset with few repos:
Screenshot_2019-06-11 Welcome

Known problem: 
- with a large number of repositories, some of the charts will timeout for the first time. Refreshing the page will re-trigger the calculation and the charts will be displayed at the end.
- with a huge number of repositories and weak hardware, some of the charts may timeout permanently


marnovo:
@smacker that's already a good baseline, what's the best workflow so that could add others and change these later?

marnovo:
Actually might be better to merge this already, and we make further the changes in a new PR.

smacker:
> what's the best workflow so that could add others and change these later?

1. run srcd-ce with this change
2. add new charts / modify current dashboard
3. export it into json
4. replace json file in the repository & PR

carlosms:
Was this one exported from the UI? I noticed that exporting from the CLI produces much smaller json files, with much less metadata.
My guess is that since we are also importing from the CLI, it will be less likely to run into problems if we also export from the CLI (`docker-compose exec sourced-ui superset export_dashboards`)

marnovo:
FYI this is the current WIP dashboard mockup with charts/queries we have added (outlined in green), all pure gitbase so far: https://docs.google.com/presentation/d/1PFUbWZiSKae2VEQclswvW02B-BlHIyNk-PeFupXXVjo/edit#slide=id.p

smacker:
@carlosms if I export it using cli I get:
Screenshot 2019-06-12 at 19 22 22

Only 2 charts work.

smacker:
Hm... It happened only after deleting & uploading. With a clean state works ok. Re-pushed.

carlosms:
Can you please take a look at the travis failing tests?

smacker:
they are failing because travis for some reason can't start postgres and does something wrong with mysql. It is unrelated to changes in this PR.
You can see the same problem in other recent prs as well.

Issue #111, opened 2019-06-14
Add dashboard for ghsync metadata


carlosms:
Part of https://github.com/src-d/ghsync/issues/24, might depend on having other tasks done.

Let's create a dashboard with all the charts defined in "List of charts we can get data quick (seconds-minutes)" from the [design doc](https://docs.google.com/document/d/1YcK11_qHTN_cH8ES_GQaVv_s1gVNQtMZNJnd6hjImJw/edit#heading=h.7ysg4ib59le1)

PR #137, opened 2019-06-18
Add dashboard based on metadata

smacker:
Only charts based on fast data.
PR changes other dashboards too to make the ids correct.

Fix: #111

Screenshot_2019-06-18 Collaboration(2)

any improvements to the charts and layouts will be done by product team.

### DIFF
--- a/srcd/dashboards/gitbase/welcome.json
+++ b/srcd/dashboards/gitbase/welcome.json
@@ -1,3948 +1,3714 @@
 {
-    "dashboards": [{
-        "__Dashboard__": {
-            "css": "",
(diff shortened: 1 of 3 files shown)


smacker:
@se7entyse7en it's not smoothed on demo instance. I just re-implemented them here using demo as the source of truth:

Screenshot 2019-06-19 at 12 26 21


taking into account that Konst was working on it for quite a while, I assume it should be like that.

se7entyse7en:
@smacker ok then, thanks for double checking it! 👍

smacker:
we can always improve it later. (actually product team can improve it later :) )

zurk:
FYI: `linear style` chart option does not work and changes nothing for this kind of plots. Where I did smoothing I did it directly in SQL query if it is not possible to smooth it in superset UI or here: 
Screen Shot 2019-06-19 at 17 30 06

Both options are not the best. But we have what we have :/

PR #144, opened 2019-06-20
Update changelog file

### LINKED ISSUE #130: Update changelog
Changelog file is out of sync since 0.2.0.

smacker:
Fix #130

Internal stuff (like fixing ci) is skipped, prs are described in terms of different features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,14 @@
 ## [Unreleased]
 <details>
   <summary>
-    Changes that have landed in master but are not yet released.
-    Click to see more.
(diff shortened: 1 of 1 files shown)


smacker:
@dpordomingo I'm not sure it makes much sense to do such splitting for this repo. It's kind of internal repository. It doesn't make much sense without srcd-ce. So it's better to write very user friendly changelog in srcd-ce and keep this repo simple.
But if you disagree please feel free to PR the change. Thanks.

dpordomingo:
Another point to consider, is: are we going to expose in sourced-ce the relevant changes that comes because sourced-ui? Or should the user visit this repo to understand the news introduced on every ui update?

Issue #155, opened 2019-06-21
Preset chart: "Number of commits per month breakdown per repo" legend taking too much space


smola:
- On medium-sized orgs (e.g. src-d, docker), the legend takes all space. It is possible to remove the legend, and the tooltip is still visible.
- So fat the tooltip works for me on Google Chrome, although it's not showing up in Firefox. I guess it's better that at least it shows up properly in Chrome?
- Also minor thing, the title should say "repository", not "repo", for consistency.

![2019-06-21-222335_2424x516_scrot](https://user-images.githubusercontent.com/23248/59949699-32265180-9474-11e9-8f3c-bbb3fb58424e.png)

gomesfernanda:
Covered by https://github.com/src-d/sourced-ui/pull/158

smacker:
pr wasn't merged and can't be merged as it is

smacker:
fixed by #160

Issue #153, opened 2019-06-21
Repository count on preset dashboard is rounded


smola:
- Repository count on preset dashboard is rounded to 10s.
- The backing data and query is actually correct.
- Can be fixed by going to chart options -> Number format -> remove the format by clicking the `x`.

gomesfernanda:
Covered by https://github.com/src-d/sourced-ui/pull/158

smacker:
pr wasn't merged and can't be merged as it is

smacker:
fixed by #160

Issue #140, opened 2019-06-20
Most active repos chart isn't sorted by value


smacker:
<img width="1460" alt="Screenshot 2019-06-20 at 12 33 55" src="https://user-images.githubusercontent.com/406916/59842894-bfc04f00-9357-11e9-9956-4eddc26b5d53.png">

gomesfernanda:
According to my empirical investigation, for some reason, the sorting on the bar chart is done according to the first metric given on the settings (instead of the total count merged + open + closed, as desired). In this case, It's sorting by `Merged` metric.

<img width="1392" alt="most_active_001" src="https://user-images.githubusercontent.com/21245928/59980855-bf32ed00-95d1-11e9-9e81-627c2dbafb39.png">

Checking what happens if we change the order of the metrics, we notice that it will sort according to the first given metric.

<img width="1390" alt="most_active_002" src="https://user-images.githubusercontent.com/21245928/59980858-bfcb8380-95d1-11e9-948c-aef097ef1b9b.png">

<img width="1393" alt="most_active_003" src="https://user-images.githubusercontent.com/21245928/59980856-bfcb8380-95d1-11e9-9744-e5ae9b389f59.png">


The query we have is:
```sql
SELECT CONCAT(repository_owner, '/', repository_name) AS repository,
DATE(created_at) as date_created,
date(merged_at) as date_merged,
date(closed_at) as date_closed
FROM pull_requests
```
And the result is a table with columns: repo, date created, date merged, date closed.

I have no answer on how to fix this without being able to choose how to sort the bar chart or how to change the query so I could select only records from the last 30 days (we currently filter this on the Superset UI).

marnovo:
@gomesfernanda maybe in the query itself we can add one line that adds up all created/merged/closed and sort by that but not display it (sorting could be in the query itself or in the UI, whatever works w/o showing it).

gomesfernanda:
What I did for now (locally) is modify the query to get the last 30 days on it:

```sql
SELECT CONCAT(repository_owner, '/', repository_name) AS repository,
COUNT(created_at) AS count_created,
COUNT(merged_at) AS count_merged,
(COUNT(closed_at) - COUNT(merged_at)) AS count_closed,
(COUNT(*) - count(closed_at)) AS count_open
FROM pull_requests
WHERE DATE(created_at) >= DATE(now()) - INTEGER '30'
GROUP BY repository
ORDER BY count_created DESC
```

[...]

Issue #115, opened 2019-06-17
Incorrect chart for releases


smacker:
<img width="1410" alt="Screenshot 2019-06-17 at 12 24 06" src="https://user-images.githubusercontent.com/406916/59597729-d5850880-90fa-11e9-9b3e-36dd9c3d024b.png">

1. It says 45 releases of borges in one day. Which doesn't look right
2. The legend is on the right, you need to scroll to it. It's different from all other our charts.

<img width="435" alt="Screenshot 2019-06-17 at 12 25 29" src="https://user-images.githubusercontent.com/406916/59597828-07966a80-90fb-11e9-877f-35b34ea36cbc.png">

1. It says 19 releases of lookout on Thu 28. Which month? Which year?

gomesfernanda:
Covered by https://github.com/src-d/sourced-ui/pull/158

smacker:
pr wasn't merged and can't be merged as it is

smacker:
fixed by #160

PR #158, opened 2019-06-23
Improve charts display on local and orgs dashboards

gomesfernanda:
This PR covers/solves:

- [Issue 153](https://github.com/src-d/sourced-ui/issues/153): Repository count on preset dashboard is rounded
- [Issue 155](https://github.com/src-d/sourced-ui/issues/155): Preset chart: "Number of commits per month breakdown per repo" legend taking too much space
- [Issue 115](https://github.com/src-d/sourced-ui/issues/115): Incorrect chart for releases
- Changes "X Axis Format" to *mmm/yy* on charts
- Releases chart:
	a) Granularity: monthly
	b) Stack repository columns
	c) Fix query to major versions `v%.%.0` and `%.%.0`
	d) Exclude 2015 filter on query
	e) Exclude legend
- Change some charts title to be shorter and more meaningful
- Divide screen 50/50 between the charts "Average time to merge" and "Breakdown by pull request outcome"

This PR does NOT cover [Issue 140](https://github.com/src-d/sourced-ui/issues/140) (Most active repos chart isn't sorted by value) and the reason is explained on the issue itself.

**Screenshots of Overview dashboard:**
welcome1_20190623
welcome2_20190623
welcome3_20190623
welcome4_20190623
welcome5_20190623

**Screenshots of Collaboration dashboard:**

collaboration1_20190623
collaboration2_20190623
collaboration3_20190623

Signed-off-by: Fernanda Gomes <fergomes@gmail.com>


marnovo:
BTW do we need to do something related to this for this PR to be properly merged?

[(from Slack)](https://src-d.slack.com/archives/GKDAJP6F9/p1561138730027500)

> Important note about dashboards!
> 
> There are lots of ids everywhere and if we mess with them everything breaks.
> It’s very very difficult to fix ids manually, I tried but there are just too many of them.
> I also see when we use cli for the export, it re-generates ids for existing dashboard in master. Most probably because there are “gaps” in ids.
> So any new dashboards based on current master will be broken. (I’ll have to redo metadata dashboard completely now :cry: )
> Cli exports all the dashboards together but we need them separately (so for local repos mode there is no metadata dashboard).
> 
> What it means to us:
> 1. We can’t work on dashboards in parallel.
> 2. After export we need to makes sure there is no “gaps”
> 3. We have to split results of export into different json files manually (no only “dashboard” object but `__SqlaTable__` as well)
> 4. DON’T MERGE ANY CHANGES TO DASHBOARDS WITHOUT TESTING IT ON EMPTY DB

se7entyse7en:
I'm testing this locally, and it seems that some query are broken. Most probably due to what @marnovo  mentioned about the export/import of dasboards that @smacker explained.

smacker:
For me charts aren't broken by incorrect datasource but at least releases charts doesn't contain any fixes.

I also checked the json. Export was done wrong.
I don't know how to fix it except redo on clean instance and export properly.

se7entyse7en:
Also having this, but I don't know what could be the reason.

Screenshot 2019-06-24 at 12 20 37

se7entyse7en:
I had the doubt that I wasn't using latest `master`. After retrying I still have [this](https://github.com/src-d/sourced-ui/pull/158#issuecomment-504933665), but not [this](https://github.com/src-d/sourced-ui/pull/158#issuecomment-504951267).

smacker:
it is done wrong. I'll re-create a new PR with the same (most probably partially) changes.

PR #159, opened 2019-06-24
Import dashboards in a fixed order

### LINKED ISSUE #156: Default dashboard to open should be 'Welcome' instead of 'Overview'
Tested on a local build after #148 was merged. 
`set_default_dashboard.py` is setting the first ID as the default. My dashboard ids are overview 1, collaboration 2, welcome 3.

carlosms:
Fix #156.

This imports the dashboards in the order defined by their IDs.
I added `sleep` because of this comment https://github.com/src-d/sourced-ui/issues/156#issuecomment-504945680. Overview was always added before any of the other 2, it does not make sense that it gets ID 3, so I thought maybe there is something running in parallel that could be causing that...

### DIFF
--- a/superset/contrib/docker/docker-init.sh
+++ b/superset/contrib/docker/docker-init.sh
@@ -43,11 +43,16 @@ if ! fabmanager list-users --app superset | grep -q $ADMIN_LOGIN; then
     superset init
 
     # Add dashboards
-    superset import_dashboards --recursive --path /home/superset/dashboards/gitbase
+    superset import_dashboards --path /home/superset/dashboards/gitbase/welcome.json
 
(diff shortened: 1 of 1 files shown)


smacker:
did it fix the issue for you? Do you have Welcome as 2 and Collaboration as 3 now?

se7entyse7en:
I tried twice and it worked:

Screenshot 2019-06-24 at 14 09 20

it's really hacky, but I don't think we have any real short-term option until we get our hands dirty with import/export.

carlosms:
> did it fix the issue for you? Do you have Welcome as 2 and Collaboration as 3 now?

Yes, same IDs in a few tries from empty DB.

PR #164, opened 2019-06-24
Add markdown text to welcome dashboards

carlosms:


superset/contrib/docker/docker-init.sh
```diff
@@ -43,17 +43,21 @@ if ! fabmanager list-users --app superset | grep -q $ADMIN_LOGIN; then
     superset init
 
     # Add dashboards
-    superset import_dashboards --path /home/superset/dashboards/gitbase/welcome.json
+    superset import_dashboards --path /home/superset/dashboards/gitbase/overview.json
 
-    # Add metadata dashboards and set welcome dashboard as a default
-    if [ ! -z "$SYNC_MODE" ]; then
+    # Add local or orgazniation dashboards
```
smacker: `orgazniation` -> `organization`

superset/contrib/docker/superset_config.py
```diff
@@ -122,7 +122,7 @@ class CeleryConfig(object):
 
 # Defaults configuration
 
-DEFAULT_DASHBOARD_ID = 2 if SYNC_MODE else 1
+DEFAULT_DASHBOARD_ID = 2 if SYNC_MODE else 2
```
smacker: `DEFAULT_DASHBOARD_ID = 2`

carlosms: true, but this way you can see where it need to be set, if it's not the same ID in the future

superset/contrib/docker/docker-init.sh
carlosms: fixed and force-pushed

se7entyse7en:
I don't know whether we want to fix for consistency even if both works. Some urls are `https://github.com/...` others are `https://www.github.com/...`.

superset/contrib/docker/superset_config.py
smacker: um. I kind of disagree. We just need to make sure that default is 2 always. But I don't think I'm going to block the PR because of it now.

carlosms:
Done in 8526f009

se7entyse7en:
@marnovo here's the screen:

Screenshot 2019-06-24 at 20 07 46

What I was referring before regarding margins/padding is that IMO (but it's a matter of tastes I guess) s something as follow looks nicer (except for the `Empty column` label that I don't know how to remove 😛):

Screenshot 2019-06-24 at 20 06 44

marnovo:
Haha just tried the same exact same thing: 3 columns, added "dividers" for the "margin/padding" and set 3-grid-column wide.

But wow, markdown rendering on superset is very bad…

Screen Shot 2019-06-24 at 8 13 35 PM

se7entyse7en:
> 3 columns, added "dividers" for the "margin/padding" and set 3-grid-column wide.

Just to make it look clearer I put the 3 cols inside a row, so that you can change the background of the entire row from transparent to white.

marnovo:
It actually gets a tad better if you `>` quote the whole markdown, since it increases the font size. But it's semantically nonsense…
Screen Shot 2019-06-24 at 8 21 00 PM

carlosms:
I propose to review only the contents to merge and release ASAP, and continue working on the looks later.

marnovo:
@carlosms agreed, just [answered on Slack](https://src-d.slack.com/archives/C7TB5NEDN/p1561401825162300?thread_ts=1561397664.155300&cid=C7TB5NEDN):

> IMO not a blocker. We probably want to make this look better but there are a few different possible ways (columns for margins, quote text, inline CSS, proper LESS styling), so I’d rather merge & test and later settle on a given solution.

dpordomingo:
Some links here points to a landing URL that has not yet been deployed (will be maybe tomorrow).
Just leaving this for @vcoisne and @marnovo consideration (info at https://github.com/src-d/landing/pull/405)

vcoisne:
Please update the URL per @dpordomingo comment. New URL is sourced.tech/community-edition

carlosms:
You mean replace 
source{d} Community Edition(https://sourced.tech/features/) with 
source{d} Community Edition(sourced.tech/community-edition), and leave the enterprise one 
source{d} Enterprise Edition(https://sourced.tech/enterprise/) as it is?

vcoisne:
Sorry about the lack of clarity @carlosms please update the URL as follow

https://sourced.tech/features/ -> https://sourced.tech/products/community-edition
https://sourced.tech/enterprise/ -> https://sourced.tech/product/enterprise-edition

Issue #126, opened 2019-06-17
Refactor init scripts


smacker:
Currently, we run python scripts from bash separately.
But each python script initializes superset which takes a considerable amount of time.
The idea is to have only one python entry point for bootstrap to speed up it.

PR #166, opened 2019-06-25
Refactor init scripts

smacker:
Fix: #126 

Use one python file instead of bash file and bunch of python scripts.
It makes bootstrap considerably faster and makes it easier to support
the code.


srcd/contrib/docker/bootstrap.py
```diff
@@ -0,0 +1,120 @@
...
+        security_manager.sync_role_definitions()
+
+        # Add dashboards
+        dashboards_root = '/home/superset/dashboards'
+        import_dashboard(dashboards_root + '/gitbase/overview.json')
+        if conf.get('SYNC_MODE'):
+            import_dashboard(dashboards_root + '/metadata/welcome.json')
+            import_dashboard(dashboards_root + '/metadata/collaboration.json')
+        else:
+            import_dashboard(dashboards_root + '/gitbase/welcome.json')
```
smacker: I removed sleeps because I think the problem was actually in Fernanda's pr. (each json contained multiple dashboards). But please test it. It works fine without sleeps for me.

superset/contrib/docker/docker-entrypoint.sh
```diff
@@ -18,7 +18,7 @@
 set -ex
 
 if [ "$SUPERSET_NO_DB_INIT" != "true" ]; then
-    source "/home/superset/docker-init.sh"
+    python bootstrap.py
```
smacker: I didn't delete `docker-init.sh` in this pr. Most probably if we are ok with this change it will be better to revert `docker-init.sh` to original superset state to avoid conflicts there.

smacker:
repushed to satisfy linter. not changes to the code itself only style.

srcd/contrib/docker/bootstrap.py
```diff
@@ -0,0 +1,119 @@
...
+    # always gitbase script to update datasource if it was changed in env var
+    dbobj = get_or_create_datasource('gitbase', conf.get('GITBASE_DATABASE_URI'),
+                                     allow_run_async=True, allow_dml=True)
+    create_datasource_tables(dbobj, conf.get('GITBASE_DB'))
+
+    # add metadata data source only in sync mode
+    if conf.get('SYNC_MODE'):
+        dbobj = get_or_create_datasource(
+            'metadata', conf.get('METADATA_DATABASE_URI'))
+
```
se7entyse7en: no need to create datasource tables for metadata db?

se7entyse7en:
@smacker I've tried and it works for me! Thanks!

srcd/contrib/docker/bootstrap.py
smacker: true. We weren't creating them when I was working on it but now we do. Will fix.

smacker: fixed

Issue #177, opened 2019-06-28
Add dashboard id param to export cli command


smacker:
Simple change in cli only. The code of export itself already supports exporting only subset of dashboards.

PR #180, opened 2019-06-28
Add dashboard-id flag to dashboard export

smacker:
Fix: #177 

tested manually.

### DIFF
--- a/superset/superset/cli.py
+++ b/superset/superset/cli.py
@@ -276,11 +276,14 @@ def import_dashboards(path, recursive):
     '--dashboard-file', '-f', default=None,
     help='Specify the the file to export to')
 @click.option(
+    '--dashboard-id', '-i', default=None,
+    help='Specify the id of dashboard to export to')
+@click.option(
     '--print_stdout', '-p', is_flag=True, default=False,
     help='Print JSON to stdout')
-def export_dashboards(print_stdout, dashboard_file):
+def export_dashboards(print_stdout, dashboard_file, dashboard_id):
     """Export dashboards to JSON"""
-    data = dashboard_import_export.export_dashboards(db.session)
+    data = dashboard_import_export.export_dashboards(db.session, dashboard_id)
     if print_stdout or not dashboard_file:
         print(data)
     if dashboard_file:
--- a/superset/superset/utils/dashboard_import_export.py
+++ b/superset/superset/utils/dashboard_import_export.py
@@ -38,12 +38,15 @@ def import_dashboards(session, data_stream, import_time=None):
     session.commit()
 
 
-def export_dashboards(session):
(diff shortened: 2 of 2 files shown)


superset/superset/cli.py
```diff
@@ -275,12 +275,15 @@ def import_dashboards(path, recursive):
 @click.option(
     '--dashboard-file', '-f', default=None,
     help='Specify the the file to export to')
+@click.option(
```
se7entyse7en: Isn't it better to support multiple ids?

superset/superset/utils/dashboard_import_export.py
```diff
@@ -38,12 +38,15 @@ def import_dashboards(session, data_stream, import_time=None):
     session.commit()
 
 
-def export_dashboards(session):
+def export_dashboards(session, id=None):
```
se7entyse7en: Better use `id_`.

superset/superset/cli.py
smacker: any real use case why would we need it? I don't mind to add multiple ids support but I don't see how it is supposed to be used.

superset/superset/utils/dashboard_import_export.py
smacker: thanks! fixed.

superset/superset/cli.py
se7entyse7en: I'm thinking that for demos and PoC we may want to export more than 1 dashboard and import them to another instance. But most probably it would be more useful to still have `n` separate json files instead of a single one. So maybe we could add a loop to produce different json files? The alternative would be to just run this command multiple times and it will work anyway.

smacker: for the demos it would be easier to use UI. It exports all selected dashboards as a single file. So if we add it here it should work in the same way.

My rationale for this change is only to allow to update initial dashboards without lots of unused data in the diff while we didn't stip it out in the UI.
I don't see how passing a single file with multiple dashboards would help us.

se7entyse7en: My idea of use-case is that we may want to export `n` dashboards from instance 1, and `m` dashboards from instance 2. And then we want to create instance 3 with some combination of dashboards from 1 and 2, and instance 4 with some other combinations. Having a 1-1 relationship between dashboard and json file I think that helps having more control. I just imagine all the dashboards that we create that are general enough to be used for multiple users/orgs as a library where we can just import what we want.

But this is a general idea, it's not a blocker for sure.

smacker: For 1-1 you just need to run this command multiple times with different ids, no?
You can even write a bash script. Taking into account that you might want to give different names to files it is a preferred way most probably.

Issue #178, opened 2019-06-28
Use non-integer identifiers for export/import


smacker:
Originally it was proposed to use names/slugs. But after research of current data structures and the code, it looks better to use remote_id field in params column instead. There is not much difference in terms of changes in the logic but smaller diff.

Required changes:
Update slices to always have remote_id = UUID
Remove remote_id update from slices import
Update dashboard import to use persistent remote_id

This changes implemented in this diff: https://github.com/src-d/sourced-ui/pull/171

PR #189, opened 2019-07-02
Remote ids as uuids

smacker:
Fix: #178 

I also tested it manually using 3 different instances of srcd-ce. (with sql update of permission directly in the db to be able to change existing charts for some cases)


smacker:
I would want to have Marvin's review here but he is on vacation for 2 or more weeks.
On another hand, it's a quite critical fix that we need rather sooner than later.
So I propose to merge it now and do the review afterward.
@carlosms wdyt?

dpordomingo:
Sorry for not being able to add much value with this issue

carlosms:
I think it's reasonable to move forward and merge. Marvin can make a review of the whole code afterwards, we can make incremental improvements if we find anything later.

Issue #192, opened 2019-07-03
Breakdown chart is broken


smacker:
<img width="702" alt="Screenshot 2019-07-03 at 16 31 11" src="https://user-images.githubusercontent.com/406916/60600201-08dcbe00-9db0-11e9-9fe7-903fa2cf155b.png">

<img width="701" alt="Screenshot 2019-07-03 at 16 32 50" src="https://user-images.githubusercontent.com/406916/60600326-37f32f80-9db0-11e9-8c77-335765ffffec.png">

Most probably because of the Typography PR.

dpordomingo:
This is already happening with our first release of `sourced-ui:v0.1.0`,
so I assume it's a problem with the original chart source-code.
![image](https://user-images.githubusercontent.com/2437584/60656432-b3f08480-9e4f-11e9-97ae-f9c4521101b1.png)

smacker:
it became worse. Before the text was in the wrong place but at least readable. Now (when it's white) it is not.

dpordomingo:
This is also `sourced-ui:v0.1.0`,

![image](https://user-images.githubusercontent.com/2437584/60657353-81478b80-9e51-11e9-991e-a08d611893ee.png)

What we can see, is that the problem goes wrong or better depending on the:
- chart size,
- the text length.

So yes, since changing the font size affect the text length, it's not the origin of the problem.

Example:
This chart is "ok"
![image](https://user-images.githubusercontent.com/2437584/60657551-d5527000-9e51-11e9-95a3-073437a1ec4a.png)

If you change its size, the labels explode
![image](https://user-images.githubusercontent.com/2437584/60657686-0763d200-9e52-11e9-86ae-a793cdc56bb2.png)

ricardobaeta:
Hi @smacker ! This issue is now fixed as you can see 👇 . It will be included on the typography-fixes PR. Once it's merged we close this. I only have two comments about this:

1) The text positioning was solved with css
2) For us to have the text proper positioned if its length is large, as you can see on the screenshots, the width of the chart itself on the dashboard needs to be increased.

**Dashboard**
![image](https://user-images.githubusercontent.com/1746146/60661442-d2f01600-9e51-11e9-87d6-e500b9afd646.png)

![image](https://user-images.githubusercontent.com/1746146/60661412-c1a70980-9e51-11e9-8075-4c315144297e.png)

[...]

Issue #186, opened 2019-06-29
UAST buttons in result tables has no margins


dpordomingo:
_Originally posted by @carlosms in https://github.com/src-d/sourced-ui/pull/139#issuecomment-506749695_

The UAST button inside the results table now overlaps one row with the next one so does not look as neat.

![image](https://user-images.githubusercontent.com/2437584/60379587-e5271a00-9a35-11e9-880b-68c4a9c9dba6.png)

ricardobaeta:
@dpordomingo This issue is already fixed as you can see 👇 . It will be included on the typography-fixes PR. We can close this when it's merged.

![image](https://user-images.githubusercontent.com/1746146/60603477-be5a4200-9dad-11e9-8f65-2c287c0dfe50.png)

carlosms:
Fixed in #196

Issue #185, opened 2019-06-29
Buttons does not keep size on action


dpordomingo:
_Originally posted by @carlosms in https://github.com/src-d/sourced-ui/pull/139#issuecomment-506673208_

I've found a missing change in button height. In this case it's a minor thing, but it means there might be other places where buttons are not consistent.

Before:
![Peek 2019-06-28 10-39](https://user-images.githubusercontent.com/1469173/60333539-3a5d2000-9991-11e9-8045-d18dc454434b.gif)

After:
![Peek 2019-06-28 10-38](https://user-images.githubusercontent.com/1469173/60333548-3df0a700-9991-11e9-8f13-dc089a9b4ee1.gif)

dpordomingo:
I think that button size (width:height) should not change on active state.

ricardobaeta:
Yes @dpordomingo I agree with you on this. It's already fixed as you can see 👇. It will be included on the typography-issues PR. Once merged we can close this.

![run-query-stop](https://user-images.githubusercontent.com/1746146/60600444-0e360a80-9da8-11e9-8697-e0db9f4d36a5.gif)

carlosms:
Fixed in #196

Issue #184, opened 2019-06-29
Extra outer scrollbar in SQL lab results


dpordomingo:
_Originally posted by @carlosms in https://github.com/src-d/sourced-ui/pull/139#issuecomment-506671899_

The SQL Lab results table has an extra outer scrollbar. 

![image](https://user-images.githubusercontent.com/2437584/60379552-469ab900-9a35-11e9-9ba2-effcd641cde6.png)

ricardobaeta:
Hi @dpordomingo ! This issue is now fixed as you can see 👇 . It will be included on the typography-fixes PR. Once it's merged we close this.

PR #196, opened 2019-07-04
Typography fixes for merged Branding PR's

ricardobaeta:
This PR fixes the identified issues on the UI, after we merged Branding typography PR https://github.com/src-d/sourced-ui/pull/139 and Adds UI Elements Branding to UI PR https://github.com/src-d/sourced-ui/pull/142

The issues solved are listed bellow:

- Buttons does not keep size on action https://github.com/src-d/sourced-ui/issues/185
- Extra outer scrollbar in SQL lab results https://github.com/src-d/sourced-ui/issues/184
- UAST tab changes some global styles https://github.com/src-d/sourced-ui/issues/183
- UAST buttons in result tables has no margins https://github.com/src-d/sourced-ui/issues/186
- Breakdown chart is broken https://github.com/src-d/sourced-ui/issues/192

Signed-off-by: Ricardo Baeta <ricardo@ricardobaeta.com>

### DIFF
--- a/srcd/superset/assets/stylesheets/less/custom-brand.less
+++ b/srcd/superset/assets/stylesheets/less/custom-brand.less
@@ -574,9 +574,20 @@ h4.panel-title {
 }
 
 .sql-toolbar .form-inline .btn.btn-sm.btn-primary:hover,
-.sql-toolbar .form-inline .btn.btn-sm.btn-primary:focus {
+.sql-toolbar .form-inline .btn.btn-sm.btn-primary:focus,
+.sql-toolbar .form-inline .btn.btn-sm.btn-primary:active {
   background-color: darken(@brand-primary, 3%);
+  color: #fff;
+  text-transform: uppercase;
+  padding: 0.75rem 3.25rem;
+  font-weight: 600;
+  font-size: 14px !important;
+  border-radius: 0.2rem;
   border: none;
+
+  i {
+    display: none;
+  }
 }
 
 .sql-toolbar .form-inline .btn.btn-sm.btn-default {
@@ -786,18 +797,20 @@ h4.panel-title {
 
 //= Tables
 
-.ReactVirtualized__Table
-  .ReactVirtualized__Grid.ReactVirtualized__Table__Grid
-  .ReactVirtualized__Grid__innerScrollContainer[style] {
-  height: 1260px !important;
-  max-height: 1260px !important;
-}
-
 .queryPane .SouthPane {
   height: 100% !important;
   .tab-pane {
     overflow-y: visible;
   }
+}
+
+.ReactVirtualized__Grid.ReactVirtualized__Table__Grid {
+  height: 350px !important;
+}
+
+.ReactVirtualized__Table {
+  width: 100%;
+  overflow: scroll;
 }
 
 //= Alerts, Success & Error Mesages
--- a/srcd/superset/assets/stylesheets/less/custom-brand.less
+++ b/srcd/superset/assets/stylesheets/less/custom-brand.less
@@ -568,7 +568,7 @@ h4.panel-title {
   font-size: 14px !important;
   border-radius: 0.2rem;
   border: none;
-  width: 124px;
+  min-width: 124px;
 
   i {
     display: none;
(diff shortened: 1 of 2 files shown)


carlosms:
The other issues listed seem to be fixed.

But the sql lab results is not working properly right now.

In this PR, when the table is too tall, the scrollbar is half hidden. If I try to scroll to the bottom of the table, I can't. Without the outer scrollbar some contents just disappear.
127 0 0 1_8088_superset_sqllab (3)

When the results table is too wide I can't see the bottom scrollbar:
127 0 0 1_8088_superset_sqllab

This is how it looks in v0.3.0-rc.1.

Too tall:
127 0 0 1_8088_superset_sqllab (2)

Too wide:
127 0 0 1_8088_superset_sqllab (1)

And in master (97fcccfa). Here you can see the bottom is just a little bit clipped, the horizontal scrollbar is half-hidden.

Too tall:
tall

Too wide:
wide

carlosms:
In case it wasn't clear in my previous comment, I think #184 is something that we can improve, but at worst is not much than an inconvenience. While the changes in this PR can be considered a bug.
I'll block this PR as it is, but I would be OK removing the changes for the results table and fixing #184 separately at some point later.

ricardobaeta:
Hi @carlosms Thanks for the prompt review! No need to block this PR. The issues with #184 are fixed by my latest commit `9efe159`. Look at 👇 

query-table-vertical-scroll

query-table-horizontal-scroll

carlosms:
After testing the last commit 9efe1591, I've noticed 2 problems.

1. There is always a horizontal scrollbar, even when the results fit.

9efe1591:
fits

v0.3.0-rc.1
fits-prev

When the results table is too tall, the vertical scroll gets hidden:
tall

2. The results table now has a fixed height that does not make use of the whole window.

In 9efe1591:
tall-prev

In v0.3.0-rc.1: See the gif used in point 1, it can be seen that the table has a fixed height and a lot of white space below it.

.

Since this doesn't look like a straight forward fix, I'd suggest to remove any related code from this PR, leave #184 as a known issue for now since it's mostly a cosmetic thing, and fix it independently later.

ricardobaeta:
Thanks again @carlosms. My latest commit `30fb5e7` removed the #184 related code, and I agree to tackle this on a further PR.

srcd/superset/assets/stylesheets/less/custom-brand.less
```diff
@@ -566,16 +568,32 @@ h4.panel-title {
   font-size: 14px !important;
   border-radius: 0.2rem;
   border: none;
+  width: 124px;
```
carlosms: If I understand this right, this is forcing the "Run query" and "Stop" to have the same width, right?
I think it's a bit risky to set a fixed width. For example for me "Run query" with 124px has a little bit of less margin on the right side.
But what is important, this width is based on the words "Run query", which can change if you set a different language. Try for example russian or brazilian portuguese, the text overflows the button:
image.

I don't know what's the better approach for this. Maybe a `min-width` will be a good compromise, it makes "Run query" and "Stop" have the same width, while allowing the button to grow if it contains longer text translations.

I think this fixed `width` is also present for other buttons in this PR, so this will probably apply in more places.

ricardobaeta: Thanks @carlos! I already pushed a commit `40f95cc` with `min-width`.

ricardobaeta:
Thanks @carlos! I already pushed a commit 40f95cc with min-width.

srcd/superset/assets/stylesheets/less/custom-brand.less
```diff
@@ -419,6 +419,7 @@ h4.panel-title {
   font-weight: 600;
   font-size: 14px !important;
   border-radius: 0.2rem;
+  width: 135px;
```
carlosms: Let's also apply min-width here please

ricardobaeta: Thanks @carlos! I already pushed a commit `9884af9` with min-width.

ricardobaeta:
Thanks @carlos! I already pushed a commit `9884af9` with min-width.

PR #197, opened 2019-07-04
Fix slices overriding on import

smacker:
The import itself worked correctly before but instead of overriding an
old chart it was always creating a new one.


smacker:
I would appreciate if somebody else would test it too

PR #201, opened 2019-07-08
Fix chart height in dashboard

### LINKED ISSUE #200: "Files missing summary" table needs to be taller
From 0.4.0-beta.1

image

carlosms:
Fix #200


carlosms:
Yes, this is the json as is it was exported by the `superset export_dashboards` command.

smacker:
there are some unnecessary objects exported because removing them isn't implemented yet

carlosms:
> there are some unnecessary objects exported because removing them isn't implemented yet

So can we mrege this PR as it is, or I should redo this json manually?

smacker:
It will be imported correctly. But for the sake of clearness, I would change this one field manually.
Fix for export is on the way. Will be ready today/tomorrow.

carlosms:
Done and force-pushed.

Issue #179, opened 2019-06-28
Implement export methods for exported entities


smacker:
Currently export is mostly just json.dumps of model instances from DB. Including all the relationships like users, roles, db configuration, etc. Import doesn’t require all the data and just skips it.

The solutions is to implement methods that would prepare objects before json.dumps by blacklisting and removing properties.

PR #203, opened 2019-07-08
Clean up export json

smacker:
Fix: #179 

UI export and cli export produce the same small json with only necessary values now.
I also updated existing dashboards by doing re-export.


smacker:
Tests weirdly don't fail for me locally when I run `ImportExportTests` suite case.
I'm investigating why.

smacker:
Something got cached and re-running tox wasn't cleaning it up. Removing DB helped.
Tests should be fixed now.

PR #205, opened 2019-07-09
Sort exported json

smacker:
Based on #203 (only last 2 commits are new)
Fix: #202


smacker:
Note: I didn't re-export local welcome dashboard here or in prev PR. That dashboard doesn't have any slices so it can be updated just the next time we change markdown.

PR #209, opened 2019-07-11
Fix hidden x-scrollbar

dpordomingo:
fix #184

I fixed the position of the resizer "slider" in the middle.
And added some margin in the bottom of the results place to reveal the horizontal scroll bar.

If approved, sql panel would look like as :point_down: 

image

PR #139, opened 2019-06-20
Branding typography

dpordomingo:
fix #134
fix #125
fix #124
fix #123
fix #122
fix #114

The new typography was introduced by https://github.com/src-d/sourced-ui/pull/106, but some days later it was reverted by https://github.com/src-d/sourced-ui/pull/128, so its changes were rollbacked.

Original changes made by https://github.com/src-d/sourced-ui/pull/106 were: 555765c25772120e699ef00164e622739a524304 450de62cee59cb1e01f453e98ded99c40e2a6559 43ba635032dabf0970c378425ec64572357a2807 a20fde3579c84f0aaf4c16943203facc85ea33ae ec5d9ed75e77bc37e786cb2a95763e19bf07ca0f 0414d65ee72f4dd9156f063459d5805938d02a74

After being reverted, an extra commit was applied on top of those ones: abb3431bd47b98bf6a6551f3a2a3069de9964672 to address the problems described by #125 #124 #123 #122

As requested by #134, I took the old work and applied on top of the current `master` in order to validate and merge it.

<details>
<summary><i>click to see more details of how the previous commits were rearranged on top of `master`</i></summary>
1. I took all the old six commits from above and rebase-squashed them into: d8b774bd <br />
2. And then applied <del>abb3431bd47b98bf6a6551f3a2a3069de9964672</del> &rarr; c828320a on top, to fix the described problems. <br />
3. Create this PR with the new two commits, containing the old and outdated commits.
</details>

Some screenshots testing it over the current `sourced v0.14.0-beta.5 `

Old view:
image

New view:
image

New `collaboration` dashboard:
image

PR #210, opened 2019-07-12
Fix hidden x-scrollbar

### LINKED ISSUE #184: Extra outer scrollbar in SQL lab results

dpordomingo:
fix #184
supersedes #209
related to https://github.com/src-d/sourced-ui/pull/196#issuecomment-508538202

Superset calculates the height of the filtered table with javascript, using hardcoded heights for tabs [[1]](https://github.com/src-d/sourced-ui/blob/master/superset/superset/assets/src/SqlLab/components/SouthPane.jsx#L32) and buttons [[2]](https://github.com/src-d/sourced-ui/blob/master/superset/superset/assets/src/SqlLab/components/ResultSet.jsx#L54), so when there is more than one line of tabs, the filtered table does not fit in the available space, appearing a new outer scroll (as it can be seen in the screenshot below) as described by #184.

With this PR, the filtered table is positioned using absolute coordinates, leaving space for all the components in the panel, no matter how much lines of tabs are opened.

If this PR is approved, the SQL Lab would be like this :point_down:

image

instead of how it was in the previous version without branding: `v0.3.0` :point_down:

image

### DIFF
--- a/srcd/superset/assets/stylesheets/less/custom-brand.less
+++ b/srcd/superset/assets/stylesheets/less/custom-brand.less
@@ -809,14 +809,34 @@ h4.panel-title {
 
 //= Layout Adjustments
 
-.SqlEditor .SouthPane {
-  margin-top: 20px;
-
-  .tab-content {
-    margin-top: 12px;
+.SqlEditor {
+  .gutter {
+    margin: 15px auto 7px !important;
   }
 
-  .ResultSetControls {
-    margin-bottom: 12px;
+  .SouthPane {
+    padding-bottom: 15px;
+
+    .tab-content > .tab-pane {
+      padding-top: 12px;
+    }
+
+    .filterable-table-container {
+      height: auto !important;
+      position: absolute;
+      top: 76px;
+      bottom: 0px;
+      left: 0;
+      right: 0;
+      overflow: auto hidden;
+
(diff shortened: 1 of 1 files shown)


dpordomingo:
Only tested in Chrome-Linux; could you please try in other supported browsers?

carlosms:
To add more context: the extra scrollbar that [I reported](https://github.com/src-d/sourced-ui/pull/139#issuecomment-506671899) is there even if there is only one row of tabs, since the hardcoded heights do not match the new css.

Which is a different problem to having more than one row of tabs.

Testing this locally, it fixes both problems.

@dpordomingo should this be contributed upstream? The problem will happen with more than one row of tabs, and they might also prefer to have css sizing instead of a hardcoded value that can be more prone to errors.

dpordomingo:
Yes, this could be contributed into superset.
Do you agree on merging, and postpone the contribution later?
As soon as it's accepted, I'll apply it here.

carlosms:
ok.

But let's first test it with firefox too at least.

dpordomingo:
I'd say that it's also ok in Firefox :point_down:
image

PR #221, opened 2019-07-21
Development environment with hot reloading

dpordomingo:
fix https://github.com/src-d/sourced-ce/issues/26
blocked by https://github.com/src-d/sourced-ce/pull/158

The current way to develop the UI is very painful:
- `make build` takes ~10 mins every time you run it after changing some CSS line (because it ruins the cached docker context)
- the official way is not clearly documented and is affected by the changes we made in superset because **source{d} CE**. It is needed extra wiring to run it.

This PR proposes a new way to run an environment for development (in MacOS and Linux), running (see [docs in the `CONTRIBUTING.md`](https://github.com/src-d/sourced-ui/blob/8f27929a16866cecb27942c0c3e8741d8f99aee0/CONTRIBUTING.md#run-sourced-ce-for-development-with-hot-reloading)):
```shell
$ make dev-prepare
$ sourced init local
```
If it is done in that way, the env takes ~2mins to be ready, and every hot-reload takes ~5sec or less.

### TL;DR

- You'll find below the rationale behind the decision made.
- Core changes in this PR
  - 1dc63ed Configure development environment with hot reloading
  - 597842d Polish `README.md`, moving info to `CONTRIBUTING.md`
- Some other things that could be done in a separated PR if you prefer so:
  - 95de302 Remove `make dev-patch`; I didn't find a use-case now that we have this, wdyt?
  - f32cd19 Restore `superset/contrib/docker/docker-init.sh` (unused since we joined init scripts)

### More info

When **incubator-superset** is ran in development mode, it's usually started with local directories bound to some internal paths. Content from volumes &mdash;in linux&mdash;, is owned by the user from the host machine.
Since the **incubator-superset** container is built to be run with `superset` user, content from volumes might be not writable because `superset` user is not the owner of those mounted dirs (as described by https://github.com/apache/incubator-superset/issues/6887).

Alternatives:

Run **incubator-superset** as `root`:
  - since it's not the expected way to run superset, it can be unexpected,
  - data which is written in the volumes (as `node_modules` or `__pycache__`) will be written as `root`, so it won't be writable from the host machine (without being `root`)

Run **incubator-superset** container as `root`, but run **incubator-superset** app with the user owning the bound data; to do so:
  1. **assumption:** all the bound data is owned by the same user.
  1. Change the UID of `superset`, to the UID of the owner of the bound data.
  1. Make sure the entrypoint is run by `superset` user, owning the original container data and the data bound to directories in the host machine.
  1. All the data written by `superset`, will use the UID of the user of the host, so it will be writable also by that user in the host.

Docker loads (by default) an extra `docker-compose.override.yml` file (docs [_multiple-compose-files_](https://docs.docker.com/compose/extends/#understanding-multiple-compose-files)), with that in mind, we can extend our official `docker-compose.yml` file with our custom config to let the user run **source{d} CE** in development mode with hot reloading.


dpordomingo:
I tested an initial version of this PR in MacOS successfully, but I need to try again the `chmod` thing.
I will try it and confirm it here when done.

dpordomingo:
@ricardobaeta this version should be way easier to run than the one you tried some weeks ago.

watcher
```diff
@@ -0,0 +1,48 @@
...
+  echo ""
+  echo -e "\e[33malternative:\e[0m"
+  echo "  run 'make set-override' first,"
+  echo "  then 'make patch' every time you change a file in 'srcd' dir"
+  echo "  and 'make clean' once you finish developing"
+  echo ""
+  exit 1
+fi
+
+make --no-print-directory apply-patch
```
smacker: you listen for `delete` event but doesn't delete anything

superset/contrib/docker/docker-entrypoint.sh
```diff
@@ -17,6 +17,15 @@
 #
 set -ex
 
+if [ "`whoami`" = "root" ] && [ -n "$LOCAL_USER" ]; then
```
smacker: could you please add comment why you do it here

srcd/contrib/docker/docker-compose.override.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.2'
+services:
+  sourced-ui:
+    user: 0:0
```
smacker: what does it do?

Makefile
```diff
@@ -7,6 +7,13 @@ SUPERSET_DIR = superset
 # Directory with custom code to copy into SUPERSET_DIR
 PATCH_SOURCE_DIR = srcd
 
+# Directory with template for docker-compose.override.yml
+OVERRIDE_TEMPLATE_PATH := $(PATCH_SOURCE_DIR)/contrib/docker/docker-compose.override.yml
+# Directory for the sourced global docker-compose.override.yml
+OVERRIDE_OUTPUT_PATH := $(HOME)/.sourced/compose-files/__active__/docker-compose.override.yml
+# Directory for the docker-compose.override.yml backup
+OVERRIDE_BACKUP_PATH := $(OVERRIDE_OUTPUT_PATH).bak
```
smacker: what is the use case for backup? do you handle the situation when `set-override` was called twice?

watcher
dpordomingo: good catch...
I don't want to process the event to remove files on the other side...
Do you think that would it be safe calling `git clean -fd $(SUPERSET_DIR)` while the developer is working there?
(I'm not used to this `git clean`, so I'm not sure if some changes in `$(SUPERSET_DIR)` could be lost because of the `git clean`)

If it is not, I'd just exclude the delete event, and document:
>_Elements deleted in `srcd` won't be deleted from `sourced`; you can stop the watcher and call it again instead._

Makefile
dpordomingo: Yes, if you call `set-override` twice, `$(OVERRIDE_BACKUP_PATH)` target is called only once, because it's a regular one (not a `.PHONY` one) (see [`Makefile::68`](https://github.com/src-d/sourced-ui/pull/221/files#diff-b67911656ef5d18c4ae36cb6741b7965R68))

srcd/contrib/docker/docker-compose.override.yml
dpordomingo: It runs the container as `root`, instead of as `superset` (the user **inside**, not the host one)

The reasoning is detailed in the PR description, section: **_"More info"_** (it is also added a link to an issue in Superset repo)

watcher
smacker: I don't feel comfortable if something can remove my work without my explicit action.
So I don't think `git clean` is a good solution here.

Keeping stuff can cause SUPER WEIRD behavior. For example, a user does `import from 'path/*'` somewhere.

Can we just call `rm` if we received the event?

superset/contrib/docker/docker-entrypoint.sh
dpordomingo: If the user is `root` ([as explained below](https://github.com/src-d/sourced-ui/pull/221#discussion_r307387875)), internal `superset` user `UID` is changed to match the user running docker `$LOCAL_USER`.
As explained in the PR desc, it is done because that way all the data written by `superset` user, will be also accessible through the volumes, by the local user, outside of the container.

srcd/contrib/docker/docker-compose.override.yml
smacker: YAML supports comments, let's add __short__ description there. Something like `set user as root inside the container to allow writing on host filesystem`

superset/contrib/docker/docker-entrypoint.sh
smacker: I mean add comment into code

dpordomingo: The problem, at least in Linux, is that the data written in a volume is owned by the user writing it, so if it is written by the local user, it won't be accessible for internal `superset` user; and the opposite: data written by internal `superset` user, won't be accessible for the local user out of the container.

This problem is avoided if both users: internal and external ones, share the same `UID`

smacker: In this case, comment would be something like `change permissions for files coming from image to allow to write data on host file system` or something else if I understood it wrong.
Just let's make it short.

watcher
dpordomingo: Well.... `import *`... _sigh_ 

I should dig more on how the `delete` event is logged (and also the `move` event, if it's different...) considering the different platforms...

What if it could be done in a separate PR, and documented in the meantime?

superset/contrib/docker/docker-entrypoint.sh
dpordomingo: would this comment be enough in your opinion @smacker ?
>_If the user running the shell is `root`, and it was provided the `UID` of the user running the host machine, it will be assigned that `UID` to the internal `superset` user, and then, the `/entrypoint.sh` will be ran as `superset` user (see more details at https://github.com/src-d/sourced-ui/pull/221)_

srcd/contrib/docker/docker-compose.override.yml
dpordomingo: well... that's not (fully) true, at least not directly. The whole sentence should be something like:
>_Set user as `root` inside the container to change `superset` internal `UID` to be the same as the host user. It will provide write access to the data from the volumes inside and outside of the container._

The problem is that the `user: 0:0` is not enough to _"allow writing on host filesystem"_.
The `user: 0:0` is (only) needed to change internal `superset` UID, and then run the `/entrypoint.sh` as `superset`, which will do provide the write access.

But if you feel comfortable with that other doc as you proposed, let's do it.

dpordomingo: docs added by d3dbe90292e4b14bf5569fe176849a0bf6edca86

watcher
dpordomingo: issue opened to be done separately https://github.com/src-d/sourced-ui/issues/225

dpordomingo:
I tested `make dev-prepare` in OSX and it's now working there thanks to 1eff7f6
- Avoid UID hack on OSX -> as described in PR desc, there's no problem with `UID` on OSX
- Replace `sed` with `awk` -> because arguments in BSD `sed` and GNU `sed` are different
- Adapt how `stat` command is formatted in BSD
- Adapt how the colors are requested in BSD and GNU `echo`

This is now ready for final validation @src-d/applications

Makefile
```diff
@@ -33,21 +63,38 @@ IS_RELEASE := $(shell echo $(VERSION) | grep -q -E '^v[[:digit:]]+\.[[:digit:]]+
...
+# Writes the proper `docker-compose.override.yml` as the sourced global override file
+$(OVERRIDE_BACKUP_PATH):
+	@mkdir -p ~/.sourced/compose-files/__active__
+	@if [ -f "$(OVERRIDE_OUTPUT_PATH)" ]; then \
+		cp $(OVERRIDE_OUTPUT_PATH) $(OVERRIDE_BACKUP_PATH); \
+	else \
+		echo "\033[33mno docker-compose.override.yml to backup\033[0m"; \
+	fi;
+
+# Prepares the development enviroment with with hot reloading
```
se7entyse7en: typo

Makefile
```diff
@@ -33,21 +63,38 @@ IS_RELEASE := $(shell echo $(VERSION) | grep -q -E '^v[[:digit:]]+\.[[:digit:]]+
...
+.PHONY: watch
+watch:
+	@DIRECTORY_TO_OBSERVE=$(PATCH_SOURCE_DIR) bash watcher
+
+# Writes the proper `docker-compose.override.yml` as the sourced global override file
+.PHONY: set-override
+set-override: $(OVERRIDE_BACKUP_PATH)
+	@awk '{ system("echo \""$$0"\"") }' $(OVERRIDE_TEMPLATE_PATH) > $(OVERRIDE_OUTPUT_PATH)
+
+# Writes the proper `docker-compose.override.yml` as the sourced global override file
```
se7entyse7en: this actually creates a backup.

superset/contrib/docker/docker-entrypoint.sh
```diff
@@ -27,8 +40,8 @@ elif [ "$SUPERSET_ENV" = "development" ]; then
     celery worker --app=superset.sql_lab:celery_app --pool=gevent -Ofair &
     # needed by superset runserver
     (cd superset/assets/ && npm ci)
-    (cd superset/assets/ && npm run dev) &
-    FLASK_ENV=development FLASK_APP=superset:app flask run -p 8088 --with-threads --reload --debugger --host=0.0.0.0
+    (cd superset/assets/ && npm run dev-server -- --host=0.0.0.0 --port=8088 --supersetPort=8081) &
```
se7entyse7en: Could you please add a comment or explain how the ports here are being used? In particular port `8081`.

watcher
```diff
@@ -0,0 +1,48 @@
...
+
+function watch_fswatch {
+  fswatch --recursive --one-event \
+    --event Created --event Updated --event Removed \
+    ${DIRECTORY_TO_OBSERVE}
+}
+
+
+if command -v inotifywait > /dev/null 2>&1; then
+  wichWatcher=inotify
```
se7entyse7en:
```suggestion
  whichWatcher=inotify
```

watcher
```diff
@@ -0,0 +1,48 @@
...
+  function watcher { watch_inotify; }
+elif command -v fswatch > /dev/null 2>&1; then
+  wichWatcher=fswatch
```
se7entyse7en:
```suggestion
  whichWatcher=fswatch
```

watcher
```diff
@@ -0,0 +1,48 @@
...
+  echo -e "\033[1;33m[alternative]\033[0m follow these steps:"
+  echo "1. run 'make set-override' first,"
+  echo "2. then 'make patch' every time you change a file in 'srcd' dir"
+  echo "3. and 'make clean' once you finish developing"
+  echo ""
+  exit 1
+fi
+
+make --no-print-directory apply-patch
+echo -e "\033[1;92mWatching for changes in 'srcd'; using '${wichWatcher}' ...\033[0m"
```
se7entyse7en:
```suggestion
echo -e "\033[1;92mWatching for changes in 'srcd'; using '${whichWatcher}' ...\033[0m"
```

watcher
se7entyse7en: > Can we just call `rm` if we received the event?

This shouldn't be hard, and it seems sensible and the most straightforward thing to do IMHO, isn't it?

dpordomingo: Each watcher returns the detected changes in a different way, so it should be parsed the output in both cases.
Since it is not critical, and it's documented, I think we can merge and improve later,

se7entyse7en: Okay! Thanks for the explanation!

superset/contrib/docker/docker-entrypoint.sh
dpordomingo: It's explained in their docs [apache/incubator-superset/CONTRIBUTING.md#building](https://github.com/apache/incubator-superset/blob/release--0.32/CONTRIBUTING.md#building) and it might be too much to explain this wiring... but let's try; wdyt about 885dad278550dda514bdfe68e74f6b59fd3e2f01?

CONTRIBUTING.md
```diff
@@ -0,0 +1,146 @@
...
+
+    # or, for repositories stored locally
+    $ sourced init local </path/to/repos>
+    ```
+
+    The first time you launch it, it will take some time to build all the UI assets,
+    you can see the progress of the build from `sourced-ui` logs (see next step).
+
+1. To see `sourced-ui` logs (with `npm` errors and such),
+    run from `sourced-ui` root directory:
```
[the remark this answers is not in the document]
dpordomingo: good catch.
No, it does not.
fixed

dpordomingo:
Thanks, @carlosms 
your suggestions were addressed by 22e22a29163ba8a168ffbb256d8a11aeedf5aa8d

Issue #213, opened 2019-07-18
Order slices/tables for import/export


smacker:
Example of changed order: https://github.com/src-d/sourced-ui/pull/212/files#diff-8f99a0ab0cdfba56f87162db3d1757fa

We need to investigate why do they change the order.
Probably we need to do something similar to https://github.com/apache/incubator-superset/pull/7728 or sort by some other value that doesn't change.

PR #224, opened 2019-07-25
Sort lists in exported dashboards

smacker:
Fix: #213

### DIFF
--- a/superset/superset/models/core.py
+++ b/superset/superset/models/core.py
@@ -657,7 +657,8 @@ class Dashboard(Model, AuditMixinNullable, ImportMixin):
             )
             # remove ids and relations (like owners, created by, slices, ...)
             copied_dashboard = dashboard.copy()
-            for slc in dashboard.slices:
+            sorted_slices = sorted(dashboard.slices, key=lambda x: (x.slice_name or ''))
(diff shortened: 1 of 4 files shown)


superset/superset/models/core.py
```diff
@@ -685,6 +686,11 @@ def export_dashboards(cls, dashboard_ids):
                 datasource_class = copied_datasource.__class__
                 for field_name in datasource_class.export_children:
                     field_val = getattr(eager_datasource, field_name).copy()
+                    sort_by = None
+                    if len(field_val) > 0:
+                        sort_by = field_val[0].export_ordering
```
se7entyse7en: Is `field_val[0]` always an object of the type for which you added `export_ordering`? So `SqlaTable`, `BaseMetric`, `BaseColumn`. And if `len(field_val) > 0` are all he elements of the same type?

se7entyse7en: I've just seen `ImportMixin` so I guess that `field_val[0]` is always an instance of a class extending `ImportMixin`.

smacker: yes. It's always an instance of a class with `ImportMixin` and always the same type in the list.

Issue #132, opened 2019-06-18
Improve LESS and CSS, test branding branch


carlosms:
This issue is to improve LESS code to make use of variables.
@ricardobaeta please add any other things that need to be reviewed, or that you need assistance with.

ricardobaeta:
Thanks @carlosms @smacker for this issue. Other than what you described, I don't have that much to add. The variables defined in `custom-brand.less` get overwritten somehow.

dpordomingo:
not a prio for next R6-CE, will resume next week

marnovo:
Check https://github.com/src-d/sourced-ui/pull/139#issuecomment-506677370

dpordomingo:
> The variables defined in `custom-brand.less` get overwritten somehow. If I redefine `@font-size-base:` for instance, it gets overwritten and I have to declare it on each element.
Variables in less are [lazily defined](http://lesscss.org/features/#variables-feature-lazy-loading), and we have:

```less
superset.less   <<-- entrypoint
  @imports index.less
  @imports vars.less

index.less
  @imports vars.less
  @imports override.less

vars.less
  @brandColor: red

override.less
  @brandColor: blue
  body {
    color: @brandColor
  }
```

When `superset.less` is compiled into `superset.css`, it will appear as it follows:


what would be equivalent to:
```less
superset.less
  @brandColor: red
  @brandColor: blue
  body {
    color: @brandColor
  }
  @brandColor: red
```

so since vars are defined twice (the second one at the end), that one wins, so no override is done:
```css
superset.css
  body {
    color: red
  }
```

dpordomingo:
Here is my diagnostic considering our goal of applying our branding over Apache Superset, and some recommendations that I'd like to address to enhance the developer experience when rebranding Apache Superset.

_**note:**_ all routes are relative to `superset/superset/assets/`

In Apache Superset there are many ways to define the APP styles; they can be summarized in two:
- `GLOBAL`: `src/theme.js`, which calls `stylesheets/superset.less`
- `COMPONENTS`: each view, component or element, has its own `.jsx` which loads its own `less` or `css`

**current limitations:**
- Both are independent, and they do not share common variables unless explicitly import them.
- Both define the same kind of properties (layout, cosmetics, typography...).
- Styles applied by `COMPONENTS` are applied **after** `GLOBAL`.

## variables definition

About how dimensions, colors, and typographies are defined, we have four situations, that are detailed in each collapsible section below.

[...]

PR #226, opened 2019-07-29
Override global variables with our branding

dpordomingo:
required by https://github.com/src-d/sourced-ui/issues/132

It solves the problem described by https://github.com/src-d/sourced-ui/issues/132#issuecomment-514847142

If it is merged, we will be able to override and use Apache Superset variables, as seen [in this example](https://github.com/src-d/sourced-ui/pull/226/files#diff-f6c2eabcf20b8e651823e72b5273008bR7).

I only detected the following three small issues, that could be solved in a separate PR:
- [ ] #229 "extra" border in dashboard headers 
- [ ] #230 Charts title is slightly bigger than the prev one 
- [ ] #231 subtitles in the charts aside are slightly different than before


dpordomingo:
# "extra" border
new: this "extra" border between main header, and subheader (on dashboards), should not exist
image

old: no "extra" border
image

dpordomingo:
# bigger title
new: "Charts" title is slightly bigger than before
image

old: original size of "Charts title"
image

dpordomingo:
# bigger subtitles in the aside
new: subtitles in the aside are slightly bigger than before
e.g.: "Datasource & Chart Type" and "Time
image

old: original size of subtitles in the aside
e.g.: "Datasource & Chart Type" and "Time
image

srcd/superset/assets/stylesheets/less/custom-brand.less
```diff
@@ -1,134 +1,10 @@
-/**
```
se7entyse7en: Why some of the variables defined here have not been moved?

dpordomingo: @se7entyse7en do you mean: "moved to [`./variables-override.less`](https://github.com/src-d/sourced-ui/pull/226/files#diff-14ccc45e35ab92af261d436268dd9b21)"?

If so, I only moved to `./variables-override.less`, those that:
- do not exist in [`./cosmo/variables.less`](https://github.com/src-d/sourced-ui/pull/226/files#diff-a25722b6f21ec29571236aee3020385c)
- do override a variable from `./cosmo/variables.less`

Why? because since we're already importing `./cosmo/variables.less`, it's not needed to override it with the very same value &rarr; it would not be clear which ones are kept equal, and which ones are overridden.

se7entyse7en: Thanks!

dpordomingo:
@ricardobaeta feel free to merge on your own, and solve the mentioned issues in separate PRs.
An alternative is that they're not real problems because I only detected them as "differences between old and new"

dpordomingo:
@ricardobaeta your suggestion was addressed and merged into `master`, many thanks for your review

dpordomingo:
I'd consider the contribution to upstream once I find the way to solve the caveats described in the causing issue.

Issue #191, opened 2019-07-03
Use correct formatter for metadata progress chart


smacker:
Currently it looks like this:
<img width="678" alt="Screenshot 2019-07-03 at 16 21 43" src="https://user-images.githubusercontent.com/406916/60599515-e4341680-9dae-11e9-8fce-20d8ab3395ba.png">

Most probably progress should be displayed as `7%` instead of this float.
Or we can remove the column completely because gitcollector chart doesn't have it.

PR #236, opened 2019-08-02
Fix formatter for metadata progress chart

smacker:
Fix: #191
The diff is big because it's the first update of the dashboard after ordering/changed_on fixes.


carlosms:
Is this fixing an existing issue, or is it a new problem that you found?

smacker:
@carlosms it's a fix for #191

Issue #194, opened 2019-07-03
Improve contributors charts


smacker:
In many projects only maintainer(s) has rights to merge. Which means for each contribution there will be +1 merge commit by a maintainer. Often it produces strange results that don't reflect the reality.

I see 2 options:
1. Remove merge commits from current charts
2. Keep current charts but add similar charts without merge commits

gomesfernanda:
I believe the first option is better to make the charts less distorted.

The query we currently have for "Number of contributors per email" and "Email wordcloud" is 

```sql
SELECT committer_email, COUNT(*) as n
FROM commits
WHERE committer_email <> 'noreply@github.com'
GROUP BY committer_email
ORDER BY n DESC

```

To exclude merge commits from the query we should exclude commits with 2 parents (that's what marks a merge commit on gitbase)

So, the new query would be

```sql
SELECT committer_email, COUNT(*) as n
FROM commits
WHERE committer_email <> 'noreply@github.com'
AND ARRAY_LENGTH(commit_parents) < 2
GROUP BY committer_email
ORDER BY n DESC
```

I analyzed the results for bblfsh and src-d repos and found out that the majority of merge commits are done under `noreply@github.com`, but this doesn't mean that the same happens for other organizations. I believe the addition of this conditional `WHERE [...] AND ARRAY_LENGTH(commit_parents) < 2` is great to avoid duplicities on commits that are both commits **and** merge commits.

smacker:
@gomesfernanda actually it is better to use author_email for this charts. The name of the chart is `Number of contributors` not `Number of committers`.

gomesfernanda:
So we should change the queries to:

For "Number of contributions per organization" and "Organization wordcloud"

```sql
SELECT REPLACE(JSON_EXTRACT(SPLIT(commit_author_email, '@'), '$[1]'),'"', '') as domain, COUNT(*) as n
FROM commits
WHERE commit_author_email LIKE '%%@%%' and commit_author_email NOT LIKE '%%@github.com'
AND ARRAY_LENGTH(commit_parents) < 2
GROUP BY domain
ORDER BY n DESC
```

For "Number of contributions per email" and "Email wordcloud"

```sql
SELECT commit_author_email, COUNT(*) as n
FROM commits
WHERE commit_author_email <> 'noreply@github.com'
AND ARRAY_LENGTH(commit_parents) < 2
GROUP BY commit_author_email
ORDER BY n DESC
```

[...]

PR #237, opened 2019-08-02
Improve contributors charts

smacker:
Fix: #194

- Update SQL queries
- The diff is huge because it's the first export after export logic
fixes


carlosms:
it's almost impossible to review what has changed. Could you please paste on a comment the queries that have changed? Or update the whole dashboard from master in a commit and the queries only in a second commit.

smacker:
created separate commit with re-export only. hope it helps.

PR #240, opened 2019-08-06
Update changelog for 0.4.0

carlosms:


smacker:
What about fixes to charts or disabled druid?

carlosms:
I trusted the current contents of the changelog to be up to date, but it seems it was not updated in every PR. I've pushed bbe0ed6f with the things you mentioned and the tab reordering.

PR #243, opened 2019-08-08
Add table is sqllab view in export

smacker:
Fix: #152 

As discussed before by hiding SQL Lab tables.

### DIFF
--- a/srcd/dashboards/gitbase/overview.json
+++ b/srcd/dashboards/gitbase/overview.json
@@ -266,6 +266,7 @@
                 "description": null,
                 "fetch_values_predicate": null,
                 "filter_select_enabled": false,
+                "is_sqllab_view": true,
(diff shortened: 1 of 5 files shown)


srcd/dashboards/gitbase/overview.json
```diff
@@ -518,6 +521,7 @@
                 "description": null,
                 "fetch_values_predicate": null,
                 "filter_select_enabled": false,
+                "is_sqllab_view": false,
```
carlosms: why is this one the only false?

smacker: because one of the charts uses default pre-registered table. Pre-registred tables should be visible.

carlosms:
let's also add an entry to the changelog, please!

smacker:
fixed

PR #244, opened 2019-08-08
Add sparksql engine spec

### LINKED ISSUE #227: Timestamp type not mapped correctly from Spark SQL to Hive
# Context

In superset, using `gsc`, I ran successfully the following query that output a date column:
```
SELECT t1.developer_id_filter,
    CAST(t1.date as TIMESTAMP) as date,
    t1.commit_count as commit_count,
    developer_identities.team_label as filter_team_label
FROM (
    SELECT developer_id_filter,
        commit_author_when as date,
        count(*) as commit_count
    FROM commits_developer_id
    WHERE developer_id_filter IN (SELECT dev_id FROM developer_teams)
    GROUP BY 1, 2) t1
JOIN developer_identities ON developer_identities.id = t1.developer_id_filter
```

However, when trying to draw a line chart from the table, 
no `Time Grain` is available and the line chart can not be compiled:

no_timegrain

# Investigation

It seems that Spark returns the date as `TIMESTAMP_TYPE` and not `TIMESTAMP`.
cf discussion on slack https://src-d.slack.com/archives/CJQ0DBAJV/p1564152651192700

We might need to include https://github.com/apache/incubator-superset/blob/master/superset/db_engine_specs/hive.py to `sparksql.py` and adapt it in order to fix the mapping between Spark SQL internal types and Hive types.

se7entyse7en:
Closes #227.

By connecting to `gsc` through `sparksql://gsc:10000/default`, sourced-ui is currently using `BaseEngineSpec` because that engine is the fallback for databases whose backend doesn't have a corresponding engine spec. In this case the backend for the `gitbase` Database is `sparksql` that doesn't have a corresponding engine spec. All the engines specs are registered [here](https://github.com/src-d/sourced-ui/blob/master/superset/superset/db_engine_specs.py#L1810).

This PR adds a `SparkSQLEngineSpec` that inherits from `HiveEngineSpec` and sets the correct engine name.

Using the `SparkSQLEngineSpec` fixes the bug #227, but broke the listing of the available views in the SQL Editor tab. These views are listed by calling the `get_views_name` name of the corresponding engine spec, but in this case, it's shadowed by the inherited classes (more details [here](https://github.com/apache/incubator-superset/issues/8010)). This has been fixed by redefining the `get_views_name` class method in `HiveEngineSpec`.

Additionally, the SparkSQL dialect doesn't support passing charset (nor presto, nor hive), but it shouldn't be required as thrift server uses `utf8` by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,11 @@
 ## [Unreleased]
 <details>
   <summary>
-    Changes that have landed in master but are not yet released.
-    Click to see more.
+
+### Fixed
+
+- Timestamp type not mapped correctly from Spark SQL to Hive ([#227](https://github.com/src-d/sourced-ui/issues/227))
+
   </summary>
 
 </details>
--- a/superset/contrib/docker/superset_config.py
+++ b/superset/contrib/docker/superset_config.py
@@ -96,11 +96,16 @@ if GITBASE_USER:
     if GITBASE_PASSWORD:
         GITBASE_AUTH += ':%s' % GITBASE_PASSWORD
     GITBASE_AUTH += '@'
-GITBASE_DATABASE_URI = '%s://%s%s:%s/%s?charset=utf8' % (GITBASE_PREFIX,
-                                                         GITBASE_AUTH,
-                                                         GITBASE_HOST,
-                                                         GITBASE_PORT,
-                                                         GITBASE_DB)
+
+GITBASE_QUERY_PARAMS = ''
+if not IS_EE:
+    GITBASE_QUERY_PARAMS = '?charset=utf8'
+GITBASE_DATABASE_URI = '%s://%s%s:%s/%s%s' % (GITBASE_PREFIX,
+                                              GITBASE_AUTH,
+                                              GITBASE_HOST,
+                                              GITBASE_PORT,
+                                              GITBASE_DB,
+                                              GITBASE_QUERY_PARAMS)
 
 SQLLAB_DEFAULT_DBID = 2  # set gitbase as default DB in SQL Lab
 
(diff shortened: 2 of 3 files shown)


carlosms:
It looks like some hive spec tests are now failing in travis

se7entyse7en:
Okay, I took a look at the test and it is expected to fail due to the made change. We have then two options:
1. fix the test,
2. implement the `get_views_name` function only in `SparkSQLEngineSpec`.

The second option also reduces our changes against upstream. What do you think is better?

carlosms:
I don't have a strong preference, both work for our purposes. Maybe 2 will make it easier to merge new changes from upstream until this lands in their repo. At the same time going with 1 can be better if it will be reused later to send a PR upstream.

I'd like to hear you opinion too @smacker.

smacker:
1. It's not clear for me why tests are failing. If you want to implement `get_view_names` only for `SparkSQLEngineSpec` the method should be defined there. If you think this change should be made in `HiveEngineSpec`, the tests must be updated.
2. I have don't have deep knowledge in hive. From my very lame view, it looks like we need to push support for sparksql, not changes in hive engine.

se7entyse7en:
> It's not clear for me why tests are failing. If you want to implement `get_view_names` only for `SparkSQLEngineSpec` the method should be defined there. If you think this change should be made in `HiveEngineSpec`, the tests must be updated.

The tests are failing because of the `get_view_names` method returning an empty list for `HiveEngineSpec` without using the passed arguments. In the test, these arguments are mocks that are not expected to be used, but with these changes, they're being used.

For our purposes we can simply put it in `SparkSQLEngineSpec`, that reason why I put it in `HiveEngineSpec` is that it is actually available also for the Hive dialect and our SparkSQL dialect uses that same function as it inherits from it.

I'll just put the method in the `SparkSQLEngineSpec` to simplify things.

se7entyse7en:
I force-pushed that changes.

PR #246, opened 2019-08-09
Fix SQL Lab tables' preview

### LINKED ISSUE #245: Support SQL Lab table previews with gitbase-spark-collector

se7entyse7en:
~Based on #244.~
Closes #245.

This fixes getting the table preview in SQL Lab for SparkSQL engine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Fixed
 
 - Timestamp type not mapped correctly from Spark SQL to Hive ([#227](https://github.com/src-d/sourced-ui/issues/227))
+- Fix tables' preview in SQL Lab for SparkSQL engine ([#246](https://github.com/src-d/sourced-ui/pull/246))
 
   </summary>
 
--- a/superset/superset/db_engine_specs.py
+++ b/superset/superset/db_engine_specs.py
@@ -1420,6 +1420,17 @@ class SparkSQLEngineSpec(HiveEngineSpec):
     def get_view_names(cls, inspector, schema):
         return BaseEngineSpec.get_view_names(inspector, schema)
 
+    @classmethod
+    def select_star(cls, my_db, table_name, engine, schema=None, limit=100,
+                    show_cols=False, indent=True, latest_partition=True,
+                    cols=None):
+        schema = None if schema == 'default' else schema
+        return super(SparkSQLEngineSpec, cls).select_star(
+            my_db, table_name, engine, schema=schema, limit=limit,
+            show_cols=show_cols, indent=indent, latest_partition=latest_partition,
+            cols=cols
+        )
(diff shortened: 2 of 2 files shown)


smacker:
Could you help me to understand why preview works for non-gitbase tables in spark?
Screenshot 2019-08-13 at 15 02 16

can this change break it?

smacker:
also you need to re-format the code to follow flake8 rules.

se7entyse7en:
mhmmm interesting, what tables are there? Is the screen coming from demo instance? Does the demo instance already include the SparkSQL engine spec?

smacker:
It has some tables from postgresql, some tables created in spark itself and some from gitbase. Looks like only gitbase tables preview doesn't work. I'm not sure where this particular table is coming from, postgresql or spark.

The screen is from demo instance.

It does not include SparkSQL engine spec.

se7entyse7en:
Do you have an easy way to create a table in Spark? I locally rested this and previews for tables of `gitbase` db (that uses sparksql) and previews for pg tables work. But I didn't test it for tables of `gitbase` db different from the default ones.

Most probably in the case in this specif case those tables have a schema that fully qualifies them correctly. For the other tables superset uses the `default` schema which is incorrect.

smacker:
Never tried but something like [CREATE TABLE](https://docs.databricks.com/spark/latest/spark-sql/language-manual/create-table.html) should work most probably.

By locally tested `pg tables` do you mean pg tables registered in gsc? It's a bit unclear from your comment.

se7entyse7en:
> By locally tested `pg tables` do you mean pg tables registered in gsc? It's a bit unclear from your comment.

Nope, sorry. Only pg tables in `main` database. I didn't tested on any table except from those available by default (`commits`, `refs`, etc.).

smacker:
let's test with manually registered tables from pg and manually created tables in spark and make sure this change doesn't break them. (or maybe discover another solution for the issue during the process)

se7entyse7en:
@smacker I tested this again with:
1. gitbase view,
2. pg table registered on spark,
3. spark table created from SQL Lab.

The preview works for all of them 🎉.

1. gitbase view (`[view] commits`)

Screenshot 2019-08-16 at 10 18 15

2. pg table registered on spark (I just registered the `slices` db)

Screenshot 2019-08-16 at 10 18 33

3. spark table created from SQL Lab (I created a simple table with 3 int cols where I added 3 rows)

Screenshot 2019-08-16 at 10 19 25

carlosms:
In case you missed it, flake8 test is failing on travis

se7entyse7en:
@carlosms Thanks, I forgot about that. BTW, travis doesn't seem to notice the changes, I'm gonna close and reopen this.

PR #247, opened 2019-08-09
Docker stop speedup using `tini`

### LINKED ISSUE #239: Exit 137 on `docker stop`
When the container is stopped, the return code is `137`. Most probably this means that `SIGTERM` is not handled by sourced-ui, so the `SIGTERM` signal sent by `docker-stop` it's actually ignored, and a kill is sent after the timeout. The reason is that in the container the process is running with `pid 1`.

Sources:
- https://success.docker.com/article/what-causes-a-container-to-exit-with-code-137
- https://hackernoon.com/my-process-became-pid-1-and-now-signals-behave-strangely-b05c52cc551c

se7entyse7en:
Closes #239.

This adds [`tini`](https://github.com/krallin/tini) to be the init process and correctly handle `SIGTERM` signal sent when `docker stop` is executed.

On my laptop `docker stop` speeded up from >10s to <1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 <details>
   <summary>
 
+### Changed
+
+- Speed-up stopping of container by using `tini` to handle `SIGTERM` ([#239](https://github.com/src-d/sourced-ui/issues/239))
+
 ### Fixed
 
 - Timestamp type not mapped correctly from Spark SQL to Hive ([#227](https://github.com/src-d/sourced-ui/issues/227))
--- a/superset/contrib/docker/Dockerfile
+++ b/superset/contrib/docker/Dockerfile
@@ -39,7 +39,12 @@ RUN apt-get install -y vim less postgresql-client redis-tools
 # https://superset.incubator.apache.org/installation.html#making-your-own-build
 # https://nodejs.org/en/download/package-manager/
 RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
-    && apt-get install -y nodejs
+        && apt-get install -y nodejs
+
+# Install tini to ensure default signal handlers for the application
+ENV TINI_VERSION v0.18.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini
 
 WORKDIR /home/superset
 
@@ -67,7 +72,8 @@ RUN cd superset/assets \
 COPY contrib/docker/superset_config.py ./superset/
 COPY contrib/docker/bootstrap.py .
 COPY contrib/docker/docker-entrypoint.sh /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
+# Remap exit code 143 (= 128 + 15 [SIGTERM]) to 0
+ENTRYPOINT ["/tini", "-e", "143", "--", "/entrypoint.sh"]
 
 HEALTHCHECK CMD ["curl", "-f", "http://localhost:8088/health"]
 


carlosms:
Looking at the docs, I see another alternative is `docker --init`, or adding `init: true` to the docker compose file.
But then probably we can't customize the exit code remap you added, correct?

Also, should we contribute this change upstream?

smacker:
I'm not sure we need tini here.
gunicorn handles sigterm correctly and actually performs a graceful-shutdown instead of just killing everybody. If you want to make it finish faster there is an option `--graceful-timeout`. But normally if there are no connections it should exit quickly.

se7entyse7en:
> But then probably we can't customize the exit code remap you added, correct?

Yup, I didn't see any way to remap the exit code, and I think that it's a nice thing to still exit with code 0.

> Also, should we contribute this change upstream?

It could make sense. Regarding the Dockerfile I was also thinking about a couple of things:
- for our use-case we may want to also have a different `Dockerfile` which is a more *production* version. For example, we don't need many of the requirements in `requirements-dev.text`. This would make the image thinner (I didn't quantify, so maybe actually not in a relevant way) and cleaner.
- I don't know very well the js ecosystem, but could it be possible/make sense to move the `npm` building stuff into another stage (multi-stage build) to speed-up the build? Now modifying even just a python requirement makes `npm` rebuild again.

se7entyse7en:
> I'm not sure we need tini here.
> gunicorn handles sigterm correctly and actually performs a graceful-shutdown instead of just killing everybody. If you want to make it finish faster there is an option --graceful-timeout. But normally if there are no connections it should exit quickly.

But we would need to handle `SIGTERM` somehow anyway, don't we? I mean if we don't use `tini`, then we would have to handle the listening to `SIGTERM` in our `entrypoint.sh`, or am I missing something?

smacker:
gunicorn listens to SIGTERM and starts graceful-shutdown. It exists with non 0 exit code because it gets killed by docker before the shutdown is finished. I'm not sure why it takes a long time (we can investigate) but still most probably you can fix it by changing `--graceful-timeout` instead of introducing `tini`.

se7entyse7en:
> gunicorn listens to SIGTERM and starts graceful-shutdown. It exists with non 0 exit code because it gets killed by docker before the shutdown is finished.

But when docker sends the `SIGTERM` the signal is sent to the entrypoint right? But if the entrypoint is not registered to handle the `SIGTERM` signal, then nothing would happens, that why I said that we should still handle it by hand.

smacker:
Do you mean because we run gunicorn from bash script it can't catch SIGTERM as if gunicorn were an entrypoint itself?

se7entyse7en:
The problem is that docker runs it with pid 1. And processes with pid 1 doesn't handle signals as normal processes unless you manually instruct them too.

smacker:
Could you please explain it a little bit more. I don't understand the problem.
I tried it with a random image that runs `gunicorn` as pid1 by manually sending sigterm:

```
$ docker run danriti/gunicorn-flask
2019-08-13 13:55:31 [1] [INFO] Starting gunicorn 0.13.4
2019-08-13 13:55:31 [1] [INFO] Listening at: http://0.0.0.0:5000 (1)
2019-08-13 13:55:31 [1] [INFO] Using worker: sync
2019-08-13 13:55:31 [8] [INFO] Booting worker with pid: 8
2019-08-13 13:55:31 [9] [INFO] Booting worker with pid: 9
2019-08-13 13:56:49 [1] [INFO] Handling signal: term
2019-08-13 13:56:50 [9] [INFO] Worker exiting (pid: 9)
2019-08-13 13:56:50 [8] [INFO] Worker exiting (pid: 8)
2019-08-13 13:56:50 [1] [INFO] Shutting down: Master
```

```
$ ps | grep gunicorn
  PID TTY           TIME CMD
23657 ttys004    0:00.28 docker run danriti/gunicorn-flask
```

```
kill -SIGTERM 23657
```

I see that the signal was correctly handled by gunicorn.

Here the log caused by `docker stop`:

```
$ docker run danriti/gunicorn-flask
2019-08-13 14:00:44 [1] [INFO] Starting gunicorn 0.13.4
2019-08-13 14:00:44 [1] [INFO] Listening at: http://0.0.0.0:5000 (1)
2019-08-13 14:00:44 [1] [INFO] Using worker: sync
2019-08-13 14:00:44 [8] [INFO] Booting worker with pid: 8
2019-08-13 14:00:44 [9] [INFO] Booting worker with pid: 9
2019-08-13 14:00:53 [1] [INFO] Handling signal: term
2019-08-13 14:00:53 [9] [INFO] Worker exiting (pid: 9)
2019-08-13 14:00:53 [8] [INFO] Worker exiting (pid: 8)
2019-08-13 14:00:53 [1] [INFO] Shutting down: Master
```

```
$ docker ps
CONTAINER ID        IMAGE                    COMMAND                  CREATED             STATUS              PORTS               NAMES
1e3d144c89cf        danriti/gunicorn-flask   "/usr/bin/gunicorn -…"   5 seconds ago       Up 3 seconds        5000/tcp            cranky_kalam
$ docker stop 1e3d144c89cf
1e3d144c89cf
```

se7entyse7en:
I actually didn't investigate in-depth, but most probably this is happening because gunicorn registers to `SIGTERM`, so even if it runs with `pid 1` it works well. But in our case we have `entrypoint.sh` which is running with `pid 1`.

smacker:
I take it as yes to my question:

> Do you mean because we run gunicorn from bash script it can't catch SIGTERM as if gunicorn were an entrypoint itself?

Then I have another question:
Would tini send the signal to gunicorn correctly taking into account it still runs by bash? 

If not we may have a problem. Most probably CE version isn't really affected but k8s deployment is. While a user is working on dashboards in case of restart superset would be just killed and changes made by a user can be potentially lost. But graceful shutdown would wait for correct commit for few milliseconds instead.

I will answer myself. Bash wouldn't forward the signal to gunicorn from tini.
The correct way to solve the issue would be to run gunicorn from bash script with `exec` or manually forward the signal. More information is here: https://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash

We may still keep tini as pid1 for that zombie processes, though I haven't seen them ever inside srcd-ui container.

se7entyse7en:
Yes, sorry for not giving you a direct answer to the question 😅.

> Would tini send the signal to gunicorn correctly taking into account it still runs by bash?

I actually assumed that it would, but you're right I don't know whether it does given that it runs by bash.

> If not we may have a problem. Most probably CE version isn't really affected but k8s deployment is. While a user is working on dashboards in case of restart superset would be just killed and changes made by a user can be potentially lost. But graceful shutdown would wait for correct commit for few milliseconds instead.

Are you talking about multi-user here right? But this would happen event without tini, doesn't it? If somebody is working on dashboards and somebody else restarts the server then the changes would get lost.

> We may still keep tini as pid1 for that zombie processes, though I haven't seen them ever inside srcd-ui container.

I don't think it would worth introducing it for this reason only, given that as you mentioned you haven't seen them.

smacker:
> Are you talking about multi-user here right?

no. I'm talking about adding tini doesn't solve the problem with not-propagation of signal to gunicorn that can cause loss of data.

Example:
1. I was working on chart, now it looks good, so I click "save" button.
2. It makes request to gunicorn that makes insert/update query to DB
3. At this moment the container was restarted using sigterm (for example in case of new deploy)
4. Tini catches it, reports all good and container gets killed
5. Killed container mean killed connection
6. Insert/update to DB wasn't committed and got rollbacked by DB as soon as connection dropped

Correct propagation of the signals would
1. Make it exit fast for a normal case
2. Make docker wait in case there is an http request from a user still in the process

se7entyse7en:
Oh ok, I get your point now. Thanks for the explanation!

What you say totally makes sense.

> Make docker wait in case there is an http request from a user still in the process

I guess that this would require some custom stopping on our side instead of using `docker stop`.

Regarding the example, it may sound silly, but couldn't it be solved by simply alerting the users before a deploy? I mean in a very large orgs with many users it could even never happen that any user is not using the dashboard.

smacker:
Sorry, I didn't get your question.
Starting gunicorn correctly by making bash script propagate signals (with or without tini) solves both problems: "faster" stop and correct shutdown without data loss.

se7entyse7en:
> Sorry, I didn't get your question.

What I mean is that you mentioned a user working on a dashboard during a deploy would lose his/her changes because of the restart of gunicorn being killed. But that would still happen independently, doesn't it? I mean if a user is modifying a dashboard even if gunicor receives that right signal then the changes would be still lost, or am I missing something?

Regarding what to do, so do you think it's better to close this and describe better the issue or to merge this as a speedup (currently gunicorn is already getting killed, but after timeout) and open a separate issue?

smacker:
> I mean if a user is modifying a dashboard even if gunicor receives that right signal then the changes would be still lost, or am I missing something?

with the right signal, gunicorn would wait for the connection to complete and data won't be lost.

> Regarding what to do, so do you think it's better to close this and describe better the issue or to merge this as a speedup (currently gunicorn is already getting killed, but after timeout) and open a separate issue?

I think better close and create a new PR that changes bash file. It would solve "speedup" problem without introducing new dependency and has other advantages described above.

smacker:
@se7entyse7en hi Marvin! Do you want to test bash only solution and send a PR? Or do you prefer me to do this?

se7entyse7en:
If you have time feel free test it! Thanks a lot @smacker!

smacker:
Ok. I'll do it on Monday then.

smacker:
Closing in favor of https://github.com/src-d/sourced-ui/pull/271

Issue #248, opened 2019-08-12
Taking the integer part of number is not in the format options


warenlg:
A lot of time, especially with `Big Number`-s, the format options suggested are not satisfying. For example, sometimes we output `48.0` or `50`, whereas the integer part should be taken here.

Some related discussions https://src-d.slack.com/archives/CJQ0DBAJV/p1564509350025500

warenlg:
Using `SMART_NUMBER` is the solution for that

PR #249, opened 2019-08-12
Add taking the integer part as a number format option

warenlg:
Signed-off-by: Waren Long <waren@sourced.tech>

Fix https://github.com/src-d/sourced-ui/issues/248


carlosms:
@warenlg would it be possible to submit this change upstream, as a PR to https://github.com/apache/incubator-superset?

warenlg:
Yes of course

smacker:
Maybe [SMART_NUMBER formatter](https://github.com/apache-superset/superset-ui/pull/109) which is [default](https://github.com/apache/incubator-superset/pull/7137) now in upstream is a better solution for this problem.

warenlg:
Yes indeed, SMART_NUMBER is doing the job, thanks

PR #251, opened 2019-08-13
Fix sorting in most active repos chart

smacker:
Fix: #140

Looks like this dashboard wasn't updated after sorting fix too.
Anyway it's better to review this PR by running it because it updated not only the sql query but columns in it and the chart.

New query is:
```
SELECT
    CONCAT(repository_owner, '/', repository_name) AS repository,
    'merged' AS mtype,
    1 AS value,
    created_at
FROM pull_requests
WHERE merged_at IS NOT NULL
UNION
SELECT
    CONCAT(repository_owner, '/', repository_name) AS repository,
    'closed' AS mtype,
    1 AS value,
    created_at
FROM pull_requests
WHERE closed_at IS NOT NULL AND merged_at IS NULL
UNION
SELECT
    CONCAT(repository_owner, '/', repository_name) AS repository,
    'opened' AS mtype,
    1 AS value,
    created_at
FROM pull_requests
WHERE closed_at IS NULL
```

Old query was:
```
SELECT CONCAT(repository_owner, '/', repository_name) AS repository, created_at, merged_at, closed_at
FROM pull_requests
```

New chart:

Screenshot 2019-08-13 at 19 43 34

### DIFF
--- a/srcd/dashboards/metadata/collaboration.json
+++ b/srcd/dashboards/metadata/collaboration.json
@@ -6,25 +6,14 @@
                 "dashboard_title": "Collaboration",
                 "description": null,
                 "json_metadata": "{\"filter_immune_slices\": [], \"timed_refresh_immune_slices\": [], \"filter_immune_slice_fields\": {}, \"expanded_slices\": {}, \"refresh_frequency\": 0, \"default_filters\": \"{}\", \"remote_id\": \"89a7262e-fad6-4dd1-bc4e-9f583376e510\"}",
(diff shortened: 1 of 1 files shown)


carlosms:
Interesting use of SUM(value) :+1: 

I'm looking at the chart options and I have a couple of suggestions:

image

1. I think we should also update the title in this PR to 'Most Active Repositories (Last 30 Days)'
2. With a row limit of 1000 the chart can grow to a point that it's just hard to read. With src-d and bblfsh orgs, 190 repos:
image
Should we limit the rows to something smaller? The chart editing only allows a predefined set of number. 10 might be too small, 50 looks ok. Or even 100. But if this grows to 1000 I think it will just be unusable.
image
If we do this I'd also prefer to change the title to '50 Most Active Repositories (Last 30 Days)'

smacker:
> Interesting use of SUM(value)

um. I read the source code of superset but it's impossible to sort graph by anything except the first metric without huge changes.

> I think we should also update the title in this PR to 'Most Active Repositories (Last 30 Days)'

👍 

> With a row limit of 1000
> If we do this I'd also prefer to change the title to '50 Most Active Repositories (Last 30 Days)'

it's not correct. Limit affects the number of rows returned by the query, not the number of bars.
With limit 50 it would produce 50/3 bars if each repository has all 3 types of prs. Or more if some repos have only 1 or 2. Which means we can't put the number in front of the title.
Also, the result might be incorrect because it sorts by `value` of a type. If the value of one type is smaller than the top 50, it would disappear with the limit. As the result it would show, for example, that repo1 has 100 open issues only, instead of 100 open, 10 closed and 10 merged.

I push-forced the update of the title (Last 30 Days) and changed the limit to 5000.

carlosms:
ok, I think I see what you mean thanks.

se7entyse7en:
We could limit the number of orgs by adding a `WHERE repository_id IN <sub_query>` where `sub_query` is essentially the same query returning the top `n` repositories sorted by `SUM(value)`.

smacker:
@se7entyse7en current query is amended by the chart's properties. 
```
WHERE created_at >= '2019-07-20 00:00:00'
  AND created_at <= '2019-08-19 00:00:00'
```
comes from the chart configuration and not for the datasource. But this where condition can't be propagated into `sub_query`. It means we will have to hard-code the same values (most probably by using `NOW()`)  in the datasource which would be rather unexpected for a user and bug prune in the future. Like changing data range in the chart would produce incorrect results.

I would prefer to keep it as it, what do you think?

se7entyse7en:
> current query is amended by the chart's properties.

do you mean the `Time range` in the chart form? If that's what you referring for you can use the macros `{{ from_dttm }}` and `{{ to_dttm }}` to refer to them also in sub-queries.

smacker:
Oh. Thanks. Fixed.

smacker:
rebased on master, had to squash commits to resolve conflicts. Re-tested locally, all fine. Merge after CI is green.

PR #241, opened 2019-08-07
[RFC] Apply our brand colors on source{d} charts

dpordomingo:
_**disclaimer**: This PR applies our branding on current source{d} charts. It is done on different levels, as it can be seen in each commit. We can use this PR to discuss the process or do it in separate and sequential PRs. The proposed palettes should be approved or amended by @src-d/product_

List of changes:
- 3c4c93c7e Add a new theme with 4 new palettes with colors from [source{d} Brand Style Guide](https://docs.google.com/presentation/d/1GSSsJ5LsjnHFHyhNLyFvXeRIpyDLMBTsv7TXVaNQe3c#slide=id.g461b1a61d5_0_54)
- 9889bf4d1 2f06d2c11 fe18682a2 Make `srcdMain` the default palette for `SUPERSET_DEFAULT` (instead of the current `bnbColors`), and use it also as the default one when creating new charts. 
- f758a821f eadfeb369 Assign our palettes on current charts, reassigning some palettes on the current charts as [described below](https://github.com/src-d/sourced-ui/pull/241#issuecomment-519027338).

The final result can also be seen in the [overview screenshot](https://github.com/src-d/sourced-ui/pull/241#issuecomment-519027495) and in the [collaboration screenshot](https://github.com/src-d/sourced-ui/pull/241#issuecomment-519027642).

### TODO

I found some lines of code in superset still using original palettes`bnbColors`, `d3Category10` and  `blue_white_yellow`, but since it does not have impact in our current dashboards (as far as I could validate in my tests, and can be seen in the screenshots below), I think that it could be addressed in a separate PR.

### Available palettes

These are the four new palettes (plus the current ones)

PR #257, opened 2019-08-15
Use SUPERSET_DEFAULT as the default palette

dpordomingo:
supersedes #241 _(This PR is the first stage, as defined by https://github.com/src-d/sourced-ui/pull/241#issuecomment-519432613)_
blocks #259

This PR prepares our charts and selectors to use the default palette: `SUPERSET_DEFAULT` instead of the currently hardcoded palettes (`bnbColors` and `blue_white_yellow`).
The colors of the palettes are kept, so this PR does not contain branding at all.

----

List of changes:
- e9366fe1f ba76c5279 Set `SUPERSET_DEFAULT` as default for Categorical and Sequential Schemes instead of the currently hardcoded `bnbColors` and `blue_white_yellow`. It will also use it as the default on ColorScheme selectors (e.g. when creating a chart)
It does not change the palettes currently configured in our charts, but the new charts if no other color is manually assigned.
- dff62ffc4 Make `PRIMARY_COLOR` fully configurable
- 4fdf8bb2d Add custom palettes and usage examples
- 368145830 Make `SUPERSET_DEFAULT` the default palette for `AnnotationLayer`
- a72e22342 Change our charts to use `SUPERSET_DEFAULT`, instead of the old `bnbColors` or `d3Category10`.


dpordomingo:
rebased adding:
- dff62ff Make PRIMARY_COLOR fully configurable
- 4fdf8bb Add custom palettes and usage examples

PR #258, opened 2019-08-15
Use reduced palettes

dpordomingo:
depends on https://github.com/src-d/sourced-ui/pull/257
depends on #259

This PR proposes simpler palettes on charts displaying only a few different data.
_e.g. when showing only merged vs unmerged, it's only needed two colors, and it might be interesting to use a different palette._

### Description of the problem
Let's consider this palette `greenDark, green, greenLight, redDark, red, redLight` which produces nice shades of colors and relations when applied to many data (which will cause circular repetition if not enough colors compared with the amount of data to represent).

But what if the same palette is assigned to a chart representing only these 2 items `passed, failed`?
With current code, colors are randomly assigned, so there is no control about which will be assigned.
Assuming that we can control the color assignment (which can be wrong considering [[1]](https://github.com/apache/incubator-superset/issues/1365T) [[2]](https://stackoverflow.com/questions/54326962/where-are-label-colour-keys-stored-in-apache-superset/54465268) [[3]](https://github.com/apache/incubator-superset/issues/2524)), let's consider that the definition order will be kept.
Then, in that example chosen colors will be `greenDark, green`, but it would be way better if it would be `green, red`, closer to which the data represents.

### proposed solution:

Choose better palettes when different metrics are less dense.

### pros:
- we can apply our main colors also in this kind of charts.
- as a lateral effect, this PR also mitigates the unexpected color change in charts, which happens when the page is refreshed.

### cons:
- not really, but yes: if the user cares about palettes, they should choose a better palette for their data; otherwise, it will be assigned the default one, which can be ok in many cases.

## Palettes Changes 

  On _**overview**_ dashboard:

  chart                                       | reduced palette
  ---                                         | ---
  Commits Evolution                           | `srcdDuo`
  Number of commits per day of the week       | `srcdDuo`
  Number of commits to HEAD per month         | `srcdDuo`
  Total number of commits per repository      | `srcdDuo`

  On _**collaboration**_ dashboard:

  chart                             | reduced palette
  ---                               | ---
  Most Active Repositories          | `srcdSix`
  Number of pull requests created   | `srcdDuo`
  Time to Merge Change Over Time    | `srcdDuo`


dpordomingo:
@ricardobaeta There is no branding in this PR, but I also added you as a reviewer to confirm that it makes sense to have this kind of reduced palettes on charts showing only a few different metrics.

carlosms:
This adds a bit of "complexity" when users create a chart, right? Instead of choosing the sourced palette, now you have to be careful to choose a palette that contains enough colors for the data you are representing.
And this might be known by us, but not by a user customizing their dashboards.

> This change also avoids charts changing its colors to quite different ones when the page is refreshed.

What if we changed the code where the colors are assigned to make it deterministic? Can this be done? It would make the dashboard consistent no matter the number of colors in the palette, which I think is desirable.

> This PR prepares the charts displaying only a few different data, to use palettes with fewer colors.
> _e.g. when showing only merged vs unmerged, it's only needed two colors, and it might be interesting to use a different palette._

This can be interesting indeed, but if we had deterministic colors assigned, we would not need reduced palettes. We could have different full sourced palettes that could apply to any chart.

se7entyse7en:
I completely agree with @carlosms.

Also, what would happen if we choose a palette with fewer colors than those required? In case that the alternative solution of changing the color assignment is too complex I'd add already all the `src<n>` palettes instead of only those for two and six colors.

dpordomingo:
I rebased over current code, so now the commit introduced by this PR is only the last one:
- b1bf0e9 Proposes better palettes when there are fewer metrics to represent.

I also rewrote the PR description in a way that I think that it answers your (totally reasonable) questions:

## TL;TR
Is it reasonable to think that we can have a way to control how the way that colors are assigned, not based on its order, but in any kind of priorities based on, e.g. the different amount of data being displayed?
Or is it more reasonable to choose a better palette when represented data has another nature?

carlosms:
I've kind of lost the conversation after editing the original description.
But still, is the (small) extra complexity needed?
The example you propose is not really related to the number of colors and metrics, it's about how `failed`/`passed` would match better with `red`/`green`.

If `green` and `greenLight` look ok next to each other in a chart with many metrics, why is it not ok on one with 2 metrics?

On the other hand I'll trust whatever @ricardobaeta prefers on this topic.

dpordomingo:
My point is that a palette containing 16 colors (to be applied sequentially) may not order them using the same priority that if they were being used for a chart using only 2 colors.

But again. This is only a RFC, trying to help @ricardobaeta when he's creating our official palette.

dpordomingo:
Rebased over @ricardobaeta palette, already merged.

Some clarifications from the discussion above:
- deterministic palettes are not that easy to implement, and they do not cover all the requirements as they were explained above,
- if the chart creator does not care about colors, it will be selected the default one, which is ok for most of the cases. If the creator prefers a more limited palette (to have more control over the applied colors), the creator can choose it.
- this does not collide nor difficult any other furder work to improve superset or D3 palettes in order to let us control better the applied colors, i.e. using a more powerful way to define only one palette with internal priorities for its colors.

If you @src-d/applications do not have any strong opinion against this, we could merge this to be released on `v0.6.0`.

dpordomingo:
Changelog at [#280(suggestion)](https://github.com/src-d/sourced-ui/pull/280/files#r324172153)

Issue #250, opened 2019-08-13
Update superset changes from latest upstream release


carlosms:

smacker:
Merging upstream isn't straight forward because upstream releases have cherry-picked commits, so the history isn't linear.

Instruction for merge written a while ago: https://gist.github.com/smacker/b8a6ccf5539ad2a209c97fe878233ac7

Superset internals were refactored in the new release so for cancelation feature it would be easy to revert current changes and apply patch from here: https://github.com/apache/incubator-superset/pull/7611

smacker:
Another pain point is a migration that exists only in our fork. It would break the order of migrations in upstream and `db upgrade` would fail. Most probably the best solution is to introduce yet another migration that would merge our current head with what has changed in upstream. See 5c76216bec29d4ba5c9754397ba62b0bd45695e3 as example.

PR #265, opened 2019-08-21
Update superset to v0.34.0rc1

smacker:
Fix: #250 

Update log is here: https://gist.github.com/smacker/b703b2ac21a76b10c76d8944030022b5 (might be useful for review)

This PR requires manual testing. I have tested it myself but could miss something.

---


se7entyse7en:
I haven't tested this yet, but I think we most probably need to have `db_engine_specs/sparksql.py`, don't we?

smacker:
@se7entyse7en I put spark definition inside `hive` file as it mostly just inherits `hive` engine. If we continue to override methods we can always move it to a separate file.

se7entyse7en:
I tried locally and there's only one thing that didn't work for me. In SQL Lab tab if I try to get the preview of the `commits` table, here's what I get:

Screenshot 2019-08-27 at 17 31 10


And this is the stacktrace:
```
sourced-ui_1    | 2019-08-27 15:29:33,574:ERROR:root:'array_type'
sourced-ui_1    | Traceback (most recent call last):
sourced-ui_1    |   File "/home/superset/superset/sql_lab.py", line 146, in get_sql_results
sourced-ui_1    |     start_time=start_time,
sourced-ui_1    |   File "/home/superset/superset/sql_lab.py", line 323, in execute_sql_statements
sourced-ui_1    |     selected_columns, data
sourced-ui_1    |   File "/home/superset/superset/db_engine_specs/presto.py", line 757, in expand_data
sourced-ui_1    |     column["name"], column["type"].lower(), all_columns
sourced-ui_1    |   File "/home/superset/superset/db_engine_specs/presto.py", line 172, in _parse_structural_column
sourced-ui_1    |     column_name, presto_type_map[field_info[1]]()
sourced-ui_1    | KeyError: 'array_type'
```

se7entyse7en:
I tried getting the preview of all tables and this seems the only one having problems.

smacker:
@se7entyse7en does it work on master? Do you think it makes sense to track this issue separately as it related to sparksql only?

se7entyse7en:
@smacker I tested on latest released version (`v0.5.0`) and it works fine.

> Do you think it makes sense to track this issue separately as it related to sparksql only?

It's totally fine for me. 👍

PR #266, opened 2019-08-22
Fix development mode in docker for macOs

smacker:
I wasn't able to run sourced-ui container in dev mode on my machine.
It took more than 10m to reach `npm ci` step with consistent close to
100% CPU load.

This commit fixes it by:
- Mounting source code with `consistency: delegated` which is the faster
mode to bind mount
- Use docker volume for `node_modules` to avoid syncing it with host
file system at all

Required changes:
- Use root user in dev mode always, not only for Linux
- Set superset user to `node_modules` volume, by default only root has
access to it
- Replace `npm ci` with `npm install` because `npm ci` removes
`node_modules` directory before starting installing dependencies. But
mounted directory can't be removed.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 - Superset updated to v0.34.0rc1 which brings many improvements and bug fixes ([#250](https://github.com/src-d/sourced-ui/issues/250))
 
+### Changed
+
+- Improve speed of development mode on macOs ([#266](https://github.com/src-d/sourced-ui/pull/266))
+
 </details>
 
 ## [v0.5.0](https://github.com/src-d/sourced-ui/releases/tag/v0.5.0) - 2019-08-20
--- a/Makefile
+++ b/Makefile
@@ -30,25 +30,18 @@ DOCKER_PUSH_LATEST ?=
 
 # Tools
 STAT := stat -c
-
-
-# Set the user as root inside of the container to have permissions to change
-# internal `superset` UID and GUID, and make it match the host user.
-# It will grant write access to the data from the volumes,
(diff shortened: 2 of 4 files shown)


smacker:
@se7entyse7en could you please validate it works for you? AFAIK you also use macOs.

se7entyse7en:
> @se7entyse7en could you please validate it works for you? AFAIK you also use macOs.

yup, I'm on OSX. Gonna test it 👍.

se7entyse7en:
I tried to test this. I built a docker image starting from this branch, and set `SUPERSET_ENV: development` in the compose file. Then I ran `make dev-prepare` separately and then restarted using sourced-ce. I tried changing the `font-size` for the `body, html` rule in `override.less`, but nothing happens and the page is not reloaded. By from the console I have this:

Screenshot 2019-08-29 at 12 06 48

What am I missing?

smacker:
1. This commit doesn't change how hot reloading works
2. You shouldn't change anything in compose file
3. I'm not sure which `override.less` you are talking about. The one in bblfsh? Then you need to open UAST tab.
4. I just re-tested with a change in `srcd/superset/assets/stylesheets/less/custom-brand.less` and
Screenshot 2019-08-29 at 12 38 16
5. I rebased on master & re-pushed so it will use new superset

se7entyse7en:
Retested it and worked flawlessly! Thanks! 👏

smacker:
Great. Merging because we have no other macOs guys in the team.

dpordomingo:
@ricardobaeta this fix should also work for you, and speed up your development in Mac.
Could you confirm that hot reloading is now working for you?

btw. Many thanks @smacker for your fix!!

PR #271, opened 2019-09-02
Use exec to start gunicorn in entrypoint.sh

### LINKED ISSUE #239: Exit 137 on `docker stop`

smacker:
Fix #239

Use exec command to start gunicorn which allows it to catch SIGTERM and
other signals correctly.

Ref: https://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash

```
$ time docker stop 00ea877ea0f8
00ea877ea0f8

real	0m2.293s
user	0m0.045s
sys	0m0.023s
```

```
srcd-z28ty2hp_sourced-ui_1     /entrypoint.sh                   Exit 0
```

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Changed
 
 - Improve speed of development mode on macOs ([#266](https://github.com/src-d/sourced-ui/pull/266))
+- Make gunicorn to catch SIGTERM correctly which speed-ups stop command and makes container exit with correct exit code ([#239](https://github.com/src-d/sourced-ui/issues/239))
 
 </details>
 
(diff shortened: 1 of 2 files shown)


carlosms:
Is this something worth proposing upstream?

smacker:
Yes: https://github.com/apache/incubator-superset/pull/8156

Issue #272, opened 2019-09-02
Unnecessary python deps in docker image


se7entyse7en:
Currently, the `Dockerfile` also installs many python dependencies that are used for development only. It may worth removing them to reduce the image size.

PR #275, opened 2019-09-03
Dockerfile optimized

### LINKED ISSUE #273: [idea] Different docker images for prod and dev
Related to #272.

This actually a more general issue that includes #272.

The current `Dockerfile` installs a lot of things that are actually not needed to just run the service. I don't know the impact, but most probably we can reduce the total size of the image considerably.

Among these there's also nodejs that is needed to compile the static files. In case of a prod docker image, would it be possible to use a multi-stage build so that in the final image we only have the required compiled static files?

se7entyse7en:
Closes #273.
This doesn't address #272.

The `Dockerfile` has been modified in order to reduce the final image size. This is achieved by using multi-stage build. There are two builders:
1. for static files,
2. for python wheels.

The first builder compiles the static files and permits the final image to avoid installing nodejs for the production version. This also improves the usage of docker caching for this step that is really heavy.

The second builder builds the python wheels to be used in the final image. This permits to use the slim image version.

The final image depends on the build arg `DEV_BUILD`: if `DEV_BUILD` is `true`, then also os dev dependencies are installed, otherwise, they're not. The Makefile provides the `build-dev` target.

The production image size has decreased from `1.76GB` to `945MB` (decrease `>800MB`, -46%).
The development image size has decreased from `1.76GB` to `1.1GB` (decrease `>650MB`, -37.5%).

---

### DIFF
--- a/superset/contrib/docker/Dockerfile
+++ b/superset/contrib/docker/Dockerfile
@@ -18,11 +18,13 @@ FROM node:12 AS static
 
 WORKDIR /home/superset
 
-COPY superset/assets superset/assets
+# Install deps and cache them
+COPY superset/assets/package*.json superset/assets/
+RUN cd superset/assets && npm ci
 
-RUN cd superset/assets \
-        && npm ci \
-        && npm run build
+# Build
+COPY superset/assets superset/assets
+RUN cd superset/assets && npm run build
 
 FROM python:3.6 AS wheels
 
--- a/superset/contrib/docker/Dockerfile
+++ b/superset/contrib/docker/Dockerfile
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-FROM node:12 AS static
+FROM node:10 AS static
 
 WORKDIR /home/superset
 
--- a/superset/contrib/docker/Dockerfile
+++ b/superset/contrib/docker/Dockerfile
@@ -14,7 +14,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-FROM python:3.6-jessie
+FROM node:12 AS static
+
+WORKDIR /home/superset
+
+COPY superset/assets superset/assets
+
+RUN cd superset/assets \
+        && npm ci \
+        && npm run build
+
+FROM python:3.6 AS wheels
+
+WORKDIR /home/superset
+
+RUN apt-get update -y
+RUN apt-get install -y libsasl2-dev
+
+COPY requirements.txt .
+COPY requirements-dev.txt .
+COPY contrib/docker/requirements-extra.txt .
+
+RUN pip install --upgrade setuptools pip \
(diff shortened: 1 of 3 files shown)


superset/contrib/docker/Dockerfile
```diff
@@ -24,48 +51,42 @@ ENV LANG=C.UTF-8 \
...
+
 COPY requirements.txt .
 COPY requirements-dev.txt .
 COPY contrib/docker/requirements-extra.txt .
 
 RUN pip install --upgrade setuptools pip \
-    && pip install -r requirements.txt -r requirements-dev.txt -r requirements-extra.txt \
-    && rm -rf /root/.cache/pip
+        && pip install --no-index --find-links=/wheels \
+        -r requirements.txt -r requirements-dev.txt -r requirements-extra.txt \
```
smacker: `requirements-dev` are needed only for dev :)
(except db drivers which we should move to another file)

se7entyse7en: yup, this doesn't address #272. The reason is that as you said we should also reorg requirements files such as whether to store our requirements in a `requirements-build.txt` or whatever where we store db drivers and pyHive. I wanted to separate this PR which is more general.

superset/contrib/docker/Dockerfile
```diff
@@ -14,7 +14,34 @@
...
 # limitations under the License.
 #
-FROM python:3.6-jessie
+FROM node:12 AS static
+
+WORKDIR /home/superset
+
+COPY superset/assets superset/assets
+
+RUN cd superset/assets \
```
smacker: > This also improves the usage of docker caching for this step that is really heavy.

if you want to improve it further you can do:
```
# install deps and cache them
COPY superset/assets/package*.json superset/assets
RUN npm ci

# build
COPY superset/assets superset/assets
RUN npm run build
```

It will use docker cache instead of downloading and installing dependencies most of the time.

superset/contrib/docker/Dockerfile
```diff
@@ -14,7 +14,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-FROM python:3.6-jessie
+FROM node:12 AS static
```
smacker: node version in build stage and in final image aren't the same. Most probably it's not important but anyway better make them match.

superset/contrib/docker/Dockerfile
se7entyse7en: Thanks! Done [here](https://github.com/src-d/sourced-ui/pull/275/commits/a5d5b61ec0768bcd4b4b909b676d077c05ab8cf0).

superset/contrib/docker/Dockerfile
se7entyse7en: Good catch! Fixed [here](https://github.com/src-d/sourced-ui/pull/275/commits/aeffcdb829bd4ff3f890497145d639ddb25cad8a).

PR #281, opened 2019-09-13
Force docker to pull FROM images when building

dpordomingo:
After fixing MariaDB issue https://github.com/src-d/sourced-ui/pull/279 we found again that `libmariadbclient18` was missing in some others mate's PCs.

Turns out that `libmariadbclient18` was removed from `debian buster` (base of `python:3.6-slim`), but when building the image with an old version of `python:3.6` and `python:3.6-slim`, it failed.

The solution appeared deleting current `python:3.6` and `python:3.6-slim` images, and building again, which pulled the proper (and new) versions of `python:3.6` and `python:3.6-slim`.

Adding `--pull` in our `docker build` commands will do it by default, to avoid building using old versions of our FROM images.

some info: https://www.databasesandlife.com/docker-build-pull-option

---

* [x] This PR contains changes that do not require a mention in the CHANGELOG file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,9 +92,9 @@ dev-prepare: set-override watch
 # Create docker image
 .PHONY: patch
 build: patch
-	docker build -t $(DOCKER_IMAGE_NAME):$(VERSION) -f superset/contrib/docker/Dockerfile $(SUPERSET_DIR)
+	docker build --pull -t $(DOCKER_IMAGE_NAME):$(VERSION) -f superset/contrib/docker/Dockerfile $(SUPERSET_DIR)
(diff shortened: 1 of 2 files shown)


se7entyse7en:
Given that `--pull` means to `Always attempt to pull a newer version of the image` this could happen again for future versions of `debian`. Isn't it better to just change the base images as follows?
- `python:3.6` -> `python:3.6-buster`
- `python:3.6-slim` -> `python:3.6-slim-buster`

dpordomingo:
Good idea @se7entyse7en 
Travis running with that locked images from bae02c5

Issue #269, opened 2019-08-28
Move celery to a separate container


se7entyse7en:
See [here](https://github.com/src-d/sourced-ui/pull/267#discussion_r318113812).

smacker:
this was done already and released

PR #285, opened 2019-09-24
Add celery mode for init the container

smacker:
This commit adds SUPERSET_ENV=celery which runs only celery worker
Current `production` environment doesn't start celery anymore

This allows to see logs and scale the workers.

The change is breaking and requires update of sourced-ce.

Ref: #269

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Make celery workers to run as separate containers ([#269](https://github.com/src-d/sourced-ui/issues/269))
+
 ### Internal
 
 - Improve docs about development mode to [run source{d} with hot reloading](./CONTRIBUTING.md#run-sourced-ce-for-development-with-hot-reloading). Previous docs didn't explain how to build the `sourced-ui` development image.
--- a/README.md
+++ b/README.md
@@ -21,35 +21,35 @@ This repository contains the code for the [`srcd/sourced-ui`](https://hub.docker
 
 You can configure the Docker image using the following environment variables:
 
-| Environment Variable  | Description                                                     |
-|-----------------------|-----------------------------------------------------------------|
-| `ADMIN_LOGIN`         | Username for the admin user                                     |
-| `ADMIN_FIRST_NAME`    | First name of the admin user                                    |
-| `ADMIN_LAST_NAME`     | Last name of the admin user                                     |
-| `ADMIN_EMAIL`         | Email of the admin user                                         |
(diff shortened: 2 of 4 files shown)


carlosms:
I guess at least `SUPERSET_ENV` needs to be updated in the [readme](https://github.com/src-d/sourced-ui/blob/master/README.md).
Probably we should also mention that you are supposed to create 2 containers, with env `production` and `celery`.
Do all the env vars now apply to both modes? Or should we have another column saying if the env var is meant for env `production`/`celery`?

What about `make build-dev` and the [hot-reloading features](https://github.com/src-d/sourced-ui/blob/master/CONTRIBUTING.md#run-sourced-ce-for-development-with-hot-reloading)? As I understand it, it will use `SUPERSET_ENV=development` and should keep working, but did you try the automation with the latest docker-compose.yml file from https://github.com/src-d/sourced-ce/pull/249?
My concern is with the automagic override of the compose file, maybe it will create the normal superset with `SUPERSET_ENV=development` plus the `sourced-ui-celery` container with `SUPERSET_ENV=celery`?

smacker:
> I guess at least SUPERSET_ENV needs to be updated in the readme.

thanks! updated

> Do all the env vars now apply to both modes? Or should we have another column saying if the env var is meant for env production/celery?

as web server and celery share the same code base I wouldn't separate them. It is true, some vars like `SYNC_MODE` or `BBLFSH_WEB_HOST` aren't needed for celery (as the first is used only in bootstrap and the second one only on bblfsh tab) but I would prefer to always have the same environment as the code is evolving and something that isn't necessary now can be used tomorrow. And such problems are a bit hard to notice as it would work in dev mode or even in production mode as long as you don't hit that particular code path that requires var inside celery worker. 

> What about make build-dev and the hot-reloading features?

thanks! I disabled new container. Celery doesn't support auto-reloading by using a flag so to implement it we will need a slightly different image with a watcher. I think it is out of the scope of this PR (and maybe isn't necessary at all).

carlosms:
What about adding somewhere that you need to create 2 containers from the same image? It might not be obvious. I'm thinking on something quick, like this, at the end of the Description section:

The image is designed to be deployed as two containers, one with `SUPERSET_ENV=production` and the other one with `SUPERSET_ENV=celery`. Take a look at the docker-compose.yml file in the [src-d/sourced-ce](https://github.com/src-d/sourced-ce) repository for more details.

smacker:
@carlosms thanks. Added.

smacker:
Merging but please review https://github.com/src-d/sourced-ce/pull/249 to make the image from master usable with srcd-ce.

PR #290, opened 2019-09-26
Bump docker-compose.override.yml version to 3.4

smacker:
We want to use it in srcd-ce to benifit from extension fields.
Different versions of file format produce an error.

---

* [x] This PR contains changes that do not require a mention in the CHANGELOG file


smacker:
imo, manual update once per year is better than maintaining code

PR #292, opened 2019-10-02
[WIP] Add support for configuring Google OAuth logins

alexpdp7:
Known issues:

* Upon registration/login you are redirected to an ugly error message.
  If you change the URL you can access the application and things are
  working, though

* User cannot access the user list!

* No documentation for the new environment variables

Signed-off-by: Alex Corcoles <alex@pdp7.net>

---

### DIFF
--- a/superset/contrib/docker/superset_config.py
+++ b/superset/contrib/docker/superset_config.py
@@ -152,3 +152,34 @@ FLASK_APP_MUTATOR = mutator
 # This parameter documented as deprecated but frontend still uses it
 # Should match `--timeout` value of gunicorn
 SUPERSET_WEBSERVER_TIMEOUT = 300
+
+OAUTH_PROVIDER = get_env_variable('OAUTH_PROVIDER', False)
+
+if OAUTH_PROVIDER:
+    from flask_appbuilder.security.manager import AUTH_OAUTH
+
+    AUTH_TYPE = AUTH_OAUTH
+    AUTH_USER_REGISTRATION = True
+    AUTH_USER_REGISTRATION_ROLE = get_env_variable('OAUTH_REGISTRATION_ROLE')
+
+    if OAUTH_PROVIDER == 'google':
+        OAUTH_PROVIDERS = [
+            {
+                'name': 'google',
+                'icon': 'fa-google',
+                'token_key': 'access_token',
+                'remote_app': {
+                    'consumer_key': get_env_variable('OAUTH_CONSUMER_KEY'),
+                    'consumer_secret': get_env_variable('OAUTH_CONSUMER_SECRET'),
+                    'base_url': 'https://www.googleapis.com/oauth2/v2/',
+                    'request_token_params': {
+                        'scope': 'email profile'
+                    },
+                    'request_token_url': None,
+                    'access_token_url': 'https://accounts.google.com/o/oauth2/token',
+                    'authorize_url': 'https://accounts.google.com/o/oauth2/auth'
+                }
+            }
+        ]
+    else:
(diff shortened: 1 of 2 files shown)


superset/contrib/docker/superset_config.py
```diff
@@ -152,3 +152,26 @@ def mutator(f):
 # This parameter documented as deprecated but frontend still uses it
 # Should match `--timeout` value of gunicorn
 SUPERSET_WEBSERVER_TIMEOUT = 300
+
+if get_env_variable('OAUTH_PROVIDER', False) == 'google':
```
smacker: maybe we should exit with an error if `OAUTH_PROVIDER` value is unknown.

superset/contrib/docker/superset_config.py
```diff
@@ -152,3 +152,26 @@ def mutator(f):
...
+    from flask_appbuilder.security.manager import AUTH_OAUTH
+
+    AUTH_TYPE = AUTH_OAUTH
+
+    OAUTH_PROVIDERS = [
+        {'name':'google', 'icon':'fa-google', 'token_key':'access_token',
+            'remote_app': {
+                'consumer_key': get_env_variable('OAUTH_CONSUMER_KEY'),
+                'consumer_secret': get_env_variable('OAUTH_CONSUMER_SECRET'),
+                'base_url':'https://www.googleapis.com/oauth2/v2/',
```
smacker: please format the code

se7entyse7en:
From [private] [slack](https://src-d.slack.com/archives/C7TB5NEDN/p1570023247201100?thread_ts=1570018104.198500&cid=C7TB5NEDN):

> IMO it makes sense to open a PR on sourced-ui pointing to your PR :thumbsup::skin-tone-4: where we could eventually add some refinements

Actually I meant issue 😅 but I wrote PR. I think that we could use this as a base and take ownership. But there's no urgency for this to be included now.

superset/contrib/docker/superset_config.py
alexpdp7: I've done a draft implementation of that, although I don't love it

superset/contrib/docker/superset_config.py
alexpdp7: It passes flake8 now (I didn't find a specific flake8 config in the repo)

alexpdp7:
> Actually I meant issue sweat_smile but I wrote PR. I think that we could use this as a base and take ownership. But there's no urgency for this to be included now.

Don't worry, it took minutes to do it. It still lacks documentation but at least you have something to start with, I guess

superset/contrib/docker/superset_config.py
smacker: https://github.com/src-d/sourced-ui/pull/292/checks?check_run_id=244877259
sourced-ui uses flake8, pylint and black.

alexpdp7: flake8 and pylint look fine for this change (although I'm not 100% familiar with pylint).

This file doesn't follow black conventions, so I'd leave it as-is, but if you want, I can introduce a commit before applying black to the existing file, then add my change with black conventions.

smacker:
another option, I can just close it, copy-paste the code into my PR and take all the praise for the implementation of the feature 😝

alexpdp7:
:D

I really don't mind. I've had success with using git format-patch / am to get out of these situations. But I'm not proud of that- I feel like I'm doing a dumb thing instead of the "proper" solution.

se7entyse7en:
hahaha

I'd vote for this BTW:

> another option, I can just close it, copy-paste the code into my PR and take all the praise for the implementation of the feature 😝

😂 

I prefer to avoid having non-properly ready things.

Issue #304, opened 2019-10-17
[bug] Markdown chart can't be added to dashboard


EgorBu:
Hello,

When you create a markdown chart (I tried with only one of them) - after saving you can't find it when you are trying to modify the dashboard.
And when you are saving chart - you can't add it to a dashboard or create new dashboard.

smacker:
For some reason, there are also 2 types of markdown charts. I tried with both none has worked.

se7entyse7en:
I reproduced it too, and also there was already an issue asking for the differences between the `markup` and `separator` viz types, but it got no anwsers: [link](https://github.com/apache/incubator-superset/issues/3939).

se7entyse7en:
Here's confirmation from maintainer that `markup` viz will be deprecated: [link](https://github.com/apache/incubator-superset/issues/6771#issuecomment-458423098).
In the meantime, we could just avoid showing them when selecting the viz type just to avoid confusion.
@smacker wdyt?

smacker:
If you don't mind, I would like to check the source code a bit later before making any decision. 

About the deprecation, almost all current charts are going to be deprecated at some point (they already marked as legacy).
You can find issues about it in superset or [superset-ui-plugins](https://github.com/apache-superset/superset-ui-plugins) repo. But new shiny charts aren't going to happen any time soon.

se7entyse7en:
> About the deprecation, almost all current charts are going to be deprecated at some point (they already marked as legacy).

Oh, I didn't know that 👍 

> If you don't mind, I would like to check the source code a bit later before making any decision.

sure! I remove myself as assignee.

smacker:
So the reason why it can't be added to dashboard is here: https://github.com/apache/incubator-superset/pull/5418#issuecomment-406134726

It makes sense as in dashboards v2 once can add markdown using the dashboard UI itself.

I propose to remove it in our fork to avoid confusion as we don't need to maintain backward compatibility for those charts.

PR #312, opened 2019-10-22
Remove markup charts

smacker:
Fix: #304

(if tests are failing, I'll fix them. It's just too slow to run them locally)

---


superset/superset/assets/src/visualizations/presets/MainPreset.js
```diff
@@ -85,8 +84,6 @@ export default class MainPreset extends Preset {
         new LineChartPlugin().configure({ key: 'line' }),
         new LineMultiChartPlugin().configure({ key: 'line_multi' }),
         new MapBoxChartPlugin().configure({ key: 'mapbox' }),
-        new MarkupChartPlugin().configure({ key: 'markup' }),
-        new MarkupChartPlugin().configure({ key: 'separator' }),
```
se7entyse7en: oh, they were really exactly the very same thing?

smacker: the difference is here: https://github.com/apache-superset/superset-ui-plugins/blob/master/packages/superset-ui-legacy-plugin-chart-markup/src/Markup.jsx#L52

se7entyse7en: thanks for poiting it out!

